PR #61, opened 2026-04-02
fix: rename_tab was silently failing — V1 protocol has no rename command

EtanHey:
## Summary
- **Root cause**: `renameTab()` in the socket client sent `rename_tab` via V1 plain-text protocol, but cmux's socket server has no such V1 command. It returned `ERROR: Unknown command 'rename_tab'` which was silently swallowed.
- **Effect**: `rename_tab` MCP tool returned `ok` but the rename never reached cmux. Sidebar title never changed.
- **Fix**: Switch from V1 to V2 JSON-RPC `tab.action` method with `action: "rename"` — this is what cmux's CLI uses internally.
- **Verified**: V2 `tab.action` returns `ok:true` and cmux sidebar updates immediately.

### Evidence
```bash
# V1 (broken - what cmuxlayer was sending):
$ echo 'rename_tab --surface surface:5 "test"' | socat - UNIX-CONNECT:cmux.sock
ERROR: Unknown command 'rename_tab'

# V2 (fixed - what cmuxlayer now sends):
$ echo '{"method":"tab.action","params":{"action":"rename","surface_id":"surface:5","title":"test"}}' | socat - cmux.sock
{"ok":true,"result":{"action":"rename","title":"test",...}}
```

## Test plan
- [x] 335/335 tests pass
- [x] TypeScript typecheck clean
- [x] Verified V2 tab.action works against real cmux socket
- [x] Updated socket client tests to verify V2 params

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

> [!NOTE]
> ### Fix `renameTab` to use V2 `tab.action` protocol instead of non-existent V1 rename command
> The V1 protocol has no `rename_tab` command, so `CmuxSocketClient.renameTab` was silently failing. The method now calls `this.call("tab.action", { action: "rename", surface_id, title, workspace_id? })` via V2. Tests are updated to assert on the V2 request shape and the mock server no longer treats `rename_tab` as a valid V1 command.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 84c1060.</sup>
> 
>

### DIFF
--- a/src/cmux-socket-client.ts
+++ b/src/cmux-socket-client.ts
@@ -194,7 +194,7 @@ export class CmuxSocketClient {
 
   /**
    * Send a V1 plain-text command. Some cmux operations (set_status,
-   * set_progress, log, rename_tab) only exist as V1 commands.
+   * set_progress, log) only exist as V1 commands.
    */
   private sendV1(command: string): Promise<string> {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


EtanHey:
@coderabbitai review

EtanHey:
@cursor @bugbot review

PR #190, opened 2026-06-25
fix(worktree): resolve repo root from workspace cwd / configurable roots (not hardcoded ~/Gits)

tmacneil:
## Problem

`spawn_agent` and `new_worktree_split` assume every repo lives at `~/Gits/<repo>` and that worktrees must live under `~/Gits` (`src/worktree.ts` `prepareWorktree`). Creating a worktree for a repo checked out anywhere else — e.g. `~/dev/<repo>` — fails before launch:

```
git -C /Users/me/Gits/esalon-admin worktree add -b cmuxlayer/… …
fatal: cannot change to '/Users/me/Gits/esalon-admin': No such file or directory
```

The repo root is derived purely from a name + a single hardcoded base, ignoring the workspace the agent is actually being spawned into (which already knows its `current_directory`).

## Fix

Resolve the source repo root with explicit precedence:

1. **Explicit** `worktree.repoRoot` (new) / `input.repoRoot`.
2. **Authoritative** — the target workspace's git top-level: `git -C <workspace.current_directory> rev-parse --show-toplevel`.
3. **Configurable search roots** — `CMUXLAYER_WORKTREE_REPO_ROOTS` (colon-separated; `~` expanded), default `["~/Gits"]`.
4. Otherwise a **clear, actionable error** instead of building a bogus `~/Gits/<repo>` path.

The worktree destination now derives from the resolved root (a sibling `<repo>.wt/<name>`), and the path-traversal guard is anchored to that base. `spawn_agent`/`new_worktree_split` pass their target `workspace` down so resolution (2) works.

## Backward compatibility

Callers that pass `homeGitsDir` explicitly (embedders, the test suite) keep their **exact prior behavior** — that branch is preserved verbatim, and the default search root remains `~/Gits`. No change for existing `~/Gits` users.

## Tests

Added to `tests/worktree.test.ts`: workspace-cwd (git-toplevel) resolution, configurable-root resolution, explicit `repoRoot` override + sibling `.wt` placement, fallback when the workspace cwd is not a work tree, and the unresolved-repo error. Full suite green (931 passed).

## Notes

This also addresses the broader inconsistency where worktree resolution, the kiro/gemini `cd ~/Gits/<repo>` launch prefixes, and external launchers each assume a different repo location — `CMUXLAYER_WORKTREE_REPO_ROOTS` is a step toward one configurable notion of where repos live. Happy to extend the same resolution to the kiro/gemini launch prefixes in a follow-up if you'd like.

---
🤖 This is a Claude comment.

> [!NOTE]
> ### Fix worktree repo root resolution to use workspace cwd and configurable roots instead of hardcoded `~/Gits`
> - Adds `resolveRepoRoot` in [worktree.ts](https://github.com/EtanHey/cmuxlayer/pull/190/files#diff-a63b9f0a6ce7c2e7f7dd324524256bcd98a8e8829d98b8b844d630fdd23ae56d) with a precedence chain: explicit `repoRoot` arg → workspace `git rev-parse --show-toplevel` → ordered `repoRoots` search dirs (default `~/Gits`) → actionable error.
> - Exposes `CMUXLAYER_WORKTREE_REPO_ROOTS` env var (colon-separated) to configure search roots; parsed in [index.ts](https://github.com/EtanHey/cmuxlayer/pull/190/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80) and passed through `CreateServerOptions`.
> - Worktree args may now include an optional `repoRoot` field to explicitly pin the source checkout.
> - Enforces a 64-character limit (`MAX_WORKTREE_ARG_LENGTH`) on the cmux `-w` launch argument when not using the legacy `homeGitsDir` path.
> - Behavioral Change: default worktree names change from `<repo>-worker-<decimal timestamp>` to `worker-<base36 timestamp>`, producing shorter paths; over-length names are now rejected before launch.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 752d03f.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Worktree setup now supports configurable repository-root discovery via environment configuration and workspace-aware resolution.
  * Worktree spawning can anchor checkout relative to the workspace’s current directory (when available), with explicit repo-root overrides.
* **Bug Fixes**
  * Improved repository-root precedence and clearer failure messaging when a git repository can’t be resolved.
  * Added validation for worktree argument length limits and updated default worktree naming for more predictable paths.
* **Tests**
  * Expanded unit tests for repo-root resolution, workspace cwd detection, override behavior, and worktree path/name constraints.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,7 +84,15 @@ async function main() {
   }
 
   const client = await createCmuxClient();
-  const server = createServer({ client });
+  const worktreeRepoRoots = process.env.CMUXLAYER_WORKTREE_REPO_ROOTS?.split(":")
+    .map((entry) => entry.trim())
+    .filter(Boolean);
+  const server = createServer({
+    client,
+    ...(worktreeRepoRoots && worktreeRepoRoots.length > 0
+      ? { worktreeRepoRoots }
+      : {}),
+  });
   const transport = new StdioServerTransport();
   await server.connect(transport);
 }
--- a/src/server.ts
+++ b/src/server.ts
@@ -681,6 +681,12 @@ export interface CreateServerOptions {
   /** Override git worktree execution/home for tests. */
   worktreeExec?: WorktreeExec;
   worktreeHomeDir?: string;
+  /**
+   * Ordered directories searched for `<root>/<repo>` when resolving a worktree's
+   * source repo and the target workspace's cwd is not a git work tree. Defaults
+   * to [`~/Gits`]. Set from CMUXLAYER_WORKTREE_REPO_ROOTS in the CLI entrypoint.
+   */
+  worktreeRepoRoots?: string[];
   /** Override control health collection (primarily for tests). */
   controlHealthCollector?: () => Promise<ControlHealth>;
   /** Periodic control health sample interval. Defaults to env or 60000ms; 0 disables. */
@@ -1016,6 +1022,12 @@ export function createServer(opts?: CreateServerOptions): McpServer {
       path: z.string().optional(),
       branch: z.string().optional(),
       base: z.string().optional(),
+      repoRoot: z
+        .string()
+        .optional()
+        .describe(
+          "Explicit source repo checkout to branch the worktree from (overrides workspace/repoRoots resolution).",
+        ),
     }),
   ]);
 
@@ -3492,10 +3504,24 @@ export function createServer(opts?: CreateServerOptions): McpServer {
       return record;
     };
 
+    const resolveWorkspaceCwd = async (
+      workspaceRef: string | undefined,
+    ): Promise<string | undefined> => {
+      if (!workspaceRef) return undefined;
+      try {
+        const { workspaces } = await client.listWorkspaces();
+        const ws = workspaces.find((w) => w.ref === workspaceRef);
+        return ws?.current_directory ?? undefined;
+      } catch {
+        return undefined;
+      }
+    };
+
     const prepareSpawnWorktree = async (
       repo: string,
       worktree: boolean | object | undefined,
       mcpProfile: McpProfile | undefined,
+      workspaceRef?: string,
     ) => {
       if (!worktree) {
         return {
@@ -3506,11 +3532,18 @@ export function createServer(opts?: CreateServerOptions): McpServer {
       }
 
       const profile = mcpProfile ?? "inherit";
+      const worktreeObj =
+        worktree && typeof worktree === "object"
+          ? (worktree as { repoRoot?: string })
+          : undefined;
       const prepared = await prepareWorktree({
         repo,
         worktree: worktree as Parameters<typeof prepareWorktree>[0]["worktree"],
         exec: opts?.worktreeExec,
         homeGitsDir: opts?.worktreeHomeDir,
+        repoRoots: opts?.worktreeRepoRoots,
+        workspaceCwd: await resolveWorkspaceCwd(workspaceRef),
+        repoRoot: worktreeObj?.repoRoot,
       });
       return {
         prepared,
@@ -3752,6 +3785,7 @@ export function createServer(opts?: CreateServerOptions): McpServer {
             args.repo,
             args.worktree,
             args.mcp_profile as McpProfile | undefined,
+            args.workspace,
           );
 
           const result = await engine.spawnAgent({
@@ -3926,6 +3960,7 @@ export function createServer(opts?: CreateServerOptions): McpServer {
             args.repo,
             args.worktree ?? true,
             args.mcp_profile as McpProfile | undefined,
+            args.workspace,
           );
           const hasPrompt = hasInlinePrompt(args.prompt);
           const result = await engine.spawnAgent({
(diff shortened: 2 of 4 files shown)


EtanHey:
**Review — fix(worktree): resolve repo root from workspace cwd / configurable roots** 🤖

Thanks @tmacneil — this is genuinely good work. The precedence chain reads cleanly, the legacy `homeGitsDir` branch is preserved verbatim, the default is `join(homedir(), "Gits")` (no hardcoded username — transferable ✅), errors are actionable, and `isGitRepo` correctly handles the linked-worktree `.git`-file case. I checked out the PR head (`752d03f`) and the worktree suite is green (12/12). I grepped the changed `src/` files for machine-specific assumptions — no hardcoded `/Users/`, usernames, or `$HOME` paths in runtime code; only JSDoc/`describe()` strings mention `~/Gits`.

Two things I'd like to resolve before merge, plus minor notes. Framing these as collaborative — the design is sound, these are refinements on the resolution edge.

---

### 🔴 High — workspace-cwd resolution trusts `git rev-parse --show-toplevel` blindly (wrong-repo + nested-worktree)

`resolveRepoRoot` returns the workspace's git top-level as the repo root without verifying it actually corresponds to the requested `repo`. Two real failure modes:

**(a) Cross-repo contamination.** If a worktree worker for repo *X* is spawned with an explicit `workspace` whose cwd is repo *Y*, it branches off *Y* but labels the dir *X*`.wt`. This is reachable from `server.ts` today: `prepareSpawnWorktree` passes `args.workspace` straight through, so an orchestrator co-locating a different-repo worktree worker into its own workspace (an encouraged pattern) triggers it. Reproduced against the PR head:

```
asked repo : esalon-admin
>> git worktree add  -C(repoRoot)=/Users/etanheyman/.r190/cmuxlayer  branch=cmuxlayer/feat  target=/Users/etanheyman/.r190/esalon-admin.wt/feat
result.path: /Users/etanheyman/.r190/esalon-admin.wt/feat
```

The launcher then runs `esalon-adminClaude -w .../esalon-admin.wt/feat` into a worktree containing **cmuxlayer** code.

**(b) Nested worktree.** When the workspace cwd is itself a *linked worktree* (`<root>/<repo>.wt/<name>` — exactly the worktree-worker cwd this PR targets), `--show-toplevel` returns the linked worktree path, not the canonical checkout. So `repoRoot` becomes the worktree, `worktreeBase = <root>/<repo>.wt`, and you get `<root>/<repo>.wt/<repo>.wt/<name2>` branched off the worktree's HEAD instead of the main repo.

**Suggested fix:** resolve the *canonical* repo root via `git -C <cwd> rev-parse --path-format=absolute --git-common-dir` (the parent of the common `.git` is the main checkout), then verify `basename(mainRoot)` matches the sanitized `repo`; on mismatch, fall through to the `repoRoots` search rather than silently trusting the workspace. That fixes both (a) and (b) and keeps the legit "spawn into the repo's own workspace" case working. (The existing tests pass because every case uses a workspace whose toplevel basename equals `repo` — a mismatch test would surface this.)

### 🟠 Medium — 64-char `-w` clamp may itself be a transferability regression

The `MAX_WORKTREE_ARG_LENGTH = 64` cap rejects when the **full** worktree path exceeds 64 (non-legacy branch). Could you confirm what cmux actually validates — the full `-w` *path*, or just the worktree *name*? It matters for transferability: on a machine with a longer home or a deeper checkout root (e.g. `/Users/some.longer.name/code/work/<repo>.wt/worker-xxxxxxxx`), a perfectly valid worktree would be rejected pre-launch — the exact "works for everyone, any location" case this PR is about. The budget math also assumes a short prefix. If cmux's limit is on the basename, clamping the full path over-rejects; if it's truly the full path, a short repro/link in the PR would lock it in. (The motivating bug in the description was a wrong-root `fatal`, not a length error, so this cap reads as a separate, somewhat speculative guard.)

### 🟡 Minor
- **Test coverage:** the suite proves the non-`~/Gits` cases well, but there's no test that with **no** `homeGitsDir`/`repoRoots`/`workspaceCwd` the default lands on `~/Gits` (the majority path we specifically want proven on both sides). Worth one test asserting the default fallback.
- **Stale doc strings:** `server.ts` `describe()` text still says worktrees go to `~/Gits/<repo>.wt/...` (lines ~3738, ~3945) — now only the default, not always true. Quick wording refresh.
- **Out of scope, correctly:** the `kiro`/`gemini` `cd ~/Gits/<repo>` launch prefixes in `agent-engine.ts` remain hardcoded — you already flagged these for a follow-up, which is the right call; noting so the transferability story stays tracked.

---

Net: solid, transferable foundation — default is `homedir()`-based and the ~/Gits majority is preserved. The High item is the one that needs a code change before merge (silent wrong-repo is a sharp edge); happy to re-review as soon as it's in. Nice work. 🙏

EtanHey:
Closing as superseded: current main's src/worktree.ts already exposes configurable `repoRoot` and `homeGitsDir` on the worktree request (worktree.ts:42-43), covering this PR's core goal (worktree base not hardcoded to ~/Gits). The remaining delta here — env-var multi-root search via CMUXLAYER_WORKTREE_REPO_ROOTS — sits on a stale v0.2.5 base and isn't worth a rebase; if multi-root auto-search is ever actually needed, re-file fresh against the current worktree.ts. (cmuxlayerLead stale-PR disposition, 2026-07-06.)

PR #326, opened 2026-07-14
fix: spawn_agent boot hardening — verify prompt delivery, report model mismatch, truthful ready/error state

jaredloman:
## Problem

`spawn_agent` can return `ok` and boot the CLI while silently failing in ways only a human watching the pane would catch:

1. The task prompt is never delivered — the pane sits at an empty prompt.
2. The `model` arg is ignored — the session launches under a different model than requested.
3. The agent registry stays `booting` with a null pid indefinitely.
4. The screen parser reports `working` for a session that is actually sitting idle (or at a context-limit banner).

## Changes

- **`agent-engine.ts`** — new `maybeAdvanceBootingAgent()`: promotes `booting → ready` only after verifying prompt delivery, detects requested-vs-actual model mismatch, and applies a stuck-boot timeout.
- **`screen-parser.ts`** — fix idle-vs-`working` misclassification; extract `isSubmitVerifiedStatus` / `screenShowsPendingInput`.
- **`agent-types.ts` / `agent-facade.ts` / `state-manager.ts`** — plumb `submit_verified` / `prompt_delivered` / `parsed_model` / `model_mismatch` through the agent record and public shape.
- **`server.ts`** — surface the new fields.
- **tests** — new/updated coverage in `sidebar-sync`, `screen-parser`, `agent-facade`, including a regression test for the ordering bug where `booting → ready` was persisted before prompt-delivery was confirmed (which reintroduced the silent-prompt-loss failure).

## Verification

`bun run typecheck` ✓ · `bun run build` ✓ · `bun run test` — 27 files / 459 tests pass.

Prompted by real-world use driving Claude Code / Codex sessions through cmuxlayer, where a swallowed spawn failure surfaced only because an operator happened to look at the pane.

> [!NOTE]
> ### Harden `spawn_agent` boot to verify prompt delivery, detect model mismatch, and report accurate ready/error state
> - Adds `submit_verified`, `prompt_delivered`, `parsed_model`, and `model_mismatch` fields to `AgentRecord` and `PublicAgent` to track boot settlement throughout the agent lifecycle.
> - `spawn_agent` now only promotes an agent to ready when prompt submission is verified; timeout errors preserve `boot_prompt_pending` for later reconciliation by the sweep.
> - `maybeMarkBootReady` sweep errors stale managed prompts that were never verified and transitions non-interactive boots to error after a new 45s `BOOT_READY_TIMEOUT_MS` timeout.
> - Adds `computeModelMismatch` to detect when the live CLI model text does not match the requested model.
> - Fixes `inferStatus` in [screen-parser.ts](https://github.com/EtanHey/cmuxlayer/pull/326/files#diff-10170f5d46ab5ce45acb8040a69928bafe44d7f458eb4af90a211557ae574e50) so Claude sessions parked at context-limit or auto-compact banners are reported as idle instead of working.
> - Behavioral Change: auto-discovered (`auto-*`) agents are no longer advanced by the boot sweep; stale stuck-booting errors are evicted using `created_at` instead of `updated_at`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 376e54f.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added boot status verification to ensure agents become ready only after successful prompt delivery.
  - Exposed prompt submission and model-mismatch status for improved visibility.
  - Added automatic detection when the running model differs from the requested model.

- **Bug Fixes**
  - Agents that fail to become interactive now retry safely and eventually report an error instead of appearing ready.
  - Improved screen parsing to correctly identify idle Claude sessions and working states.

### DIFF
--- a/src/agent-engine.ts
+++ b/src/agent-engine.ts
@@ -1570,8 +1570,14 @@ export class AgentEngine {
       const evidence = this.readReadyEvidence(agent, screen.text);
       const hasTargetEvidence =
         evidence.ready || (targetState === "ready" && evidence.activeCodex);
+      const awaitingManagedBootPrompt =
+        targetState === "ready" &&
+        agent.boot_prompt_pending === true &&
+        agent.prompt_delivered === false &&
+        !evidence.activeCodex;
       if (
         !hasTargetEvidence ||
+        awaitingManagedBootPrompt ||
         (targetState === "ready" &&
           !evidence.activeCodex &&
           this.screenShowsPendingBootPrompt(agent, screen.text))
@@ -2595,8 +2601,12 @@ export class AgentEngine {
         agent.boot_prompt_pending === true &&
         !evidence.activeCodex &&
         this.screenShowsPendingBootPrompt(agent, screen.text);
+      const awaitingManagedBootPrompt =
+        agent.boot_prompt_pending === true &&
+        agent.prompt_delivered === false &&
+        !evidence.activeCodex;
 
-      if (agent.boot_prompt_pending && promptStillPending) {
+      if (promptStillPending || awaitingManagedBootPrompt) {
         this.readyPatternMatches.delete(agent.agent_id);
         if (this.isBootPromptPendingStale(agent)) {
           const failedSettlement = this.stateMgr.updateRecord(agent.agent_id, {
--- a/src/agent-engine.ts
+++ b/src/agent-engine.ts
@@ -403,6 +403,7 @@ const DEFAULT_STOP_POST_CONDITION_TIMEOUT_MS = 1_000;
 const STOP_POST_CONDITION_POLL_MS = 50;
 const BOOT_SESSION_CAPTURE_LINES = 80;
 const MAX_DEFERRED_TRANSCRIPT_CAPTURE_ATTEMPTS = 3;
+const BOOT_READY_TIMEOUT_MS = 45_000;
 const BOOT_PROMPT_PENDING_STALE_MS = 5 * 60_000;
 const TASK_DONE_CONFIRMATION_MS = 5_000;
 const DONE_QUIESCENCE_MS = 1_500;
@@ -439,6 +440,20 @@ function toParsedScreenStatus(
     default:
       return null;
   }
+}
+
+/**
+ * Loosely compare the requested model with the model reported by the live CLI.
+ * A missing side is unknown rather than a match.
+ */
+export function computeModelMismatch(
+  requestedModel: string,
+  parsedModel: string | null,
+): boolean | null {
+  const requested = requestedModel.toLowerCase().trim();
+  const parsed = parsedModel?.toLowerCase().trim();
+  if (!requested || !parsed) return null;
+  return !parsed.includes(requested) && !requested.includes(parsed);
 }
 
 export { buildResumeCommand } from "./agent-command.js";
@@ -1577,9 +1592,21 @@ export class AgentEngine {
               screen: Promise.resolve(screen),
             })
           : agent;
-      if (targetState === "ready" && transitionAgent.boot_prompt_pending) {
+      if (targetState === "ready") {
+        const parsedModel = parseScreen(screen.text).model;
         transitionAgent = this.stateMgr.updateRecord(transitionAgent.agent_id, {
-          boot_prompt_pending: false,
+          parsed_model: parsedModel,
+          model_mismatch: computeModelMismatch(
+            transitionAgent.model,
+            parsedModel,
+          ),
+          ...(transitionAgent.boot_prompt_pending
+            ? {
+                boot_prompt_pending: false,
+                prompt_delivered: true,
+                submit_verified: true,
+              }
+            : {}),
         });
         this.registry.set(transitionAgent.agent_id, transitionAgent);
       }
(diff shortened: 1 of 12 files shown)


EtanHey:
@jaredloman — status on this PR, and an apology for the silence: it has been open 19 days without the review it was promised. That is on us, not you.

**Where it stands** (2026-08-03 fleet PR sweep): `ahead=4 / behind=523`, currently **CONFLICTING**. Two later PRs — #343 (`fix(relay): reject false submit receipts`) and #347 (`fix: enforce ratified two-role placement`) — have since touched the same spawn/placement surface, and #347 shipped in cmuxlayer 0.4.19. Your `spawn_agent` boot-hardening premise is still correct and still wanted: verify prompt delivery, report model mismatch, truthful ready/error state. We hit exactly that class of bug repeatedly today — `boot_prompt_submit_verified: true` returned on prompts that never submitted, three separate times.

**What would help:** a rebase onto current `main` so the conflicts resolve against the shipped placement contract, then we route a review immediately — not another 19 days.

**If you would rather not carry it forward**, say so and we will close it with your context recorded and re-implement the idea natively, crediting the PR. Either is fine; what is not fine is leaving it open indefinitely with no answer.

— `@maintenance`, taking every stale PR to a terminal state

EtanHey:
Maintainer rebase completed for Jared — this was wanted contributor work, so I carried the drift instead of sending 20 days of repository churn back to him.

The branch is now rebased onto current `main` (`cb2fc78`). All four original commits still have `Jared Loman <jaredloman@MiniBot.local>` as `Author:`; I added one separate maintainer-authored hardening commit for integration behavior exposed during the rebase.

The diagnosis deserves explicit credit: on 2026-07-14 Jared named all four defect classes that burned us again 20 days later — missing task delivery, ignored model selection, indefinitely booting registry rows, and idle panes reported as working.

### Per-problem disposition

1. **Still needed — landed his.** PR #345 made initial delivery routing much stronger, but current `main` could still clear a stale pending marker and promote a live pane to `ready` without verified submission. The rebased settlement path now keeps unverified managed prompts booting, records `prompt_delivered:false` / `submit_verified:false`, and eventually errors instead of false-greening (`src/agent-engine.ts:2580`; `src/server.ts:8868`).
2. **Still needed — landed his as complementary detection.** Jared's `parsed_model` / `model_mismatch` fields make a live mismatch visible to callers (`src/agent-engine.ts:2592`; `src/agent-facade.ts:29`). This does **not** beat or replace PR #351: #351 validates aliases and fixes launcher argv/effort before spawn, while #326 observes the banner after launch. The right ruling is to keep #351 and retain this post-launch alarm.
3. **Still needed — landed his.** A managed pane that never becomes interactive now leaves `booting` after a bounded timeout with an explicit error, and forced resync can still evict that exact no-CLI ghost (`src/agent-engine.ts:2646`; `src/agent-registry.ts:2155`).
4. **Both, and HIS IS BETTER — kept his broader case.** PR #177 had already fixed the persistent `bypass permissions on` footer on current `main` (`src/screen-parser.ts:1193-1198`), so that duplicate hunk was not replayed. Jared also handled context-low/auto-compact banners that coexist with `esc to interrupt`; that precedence case was still missing and is now covered (`src/screen-parser.ts:239-243,1160`).

### Tests and rebase adaptations

- Fresh REDs on current `main` proved: a stale visible composer prompt became `ready`; a stale noninteractive shell remained `booting`; an auto-compact banner parsed `working`; and the public projection dropped submit/model evidence.
- A further adversarial RED caught an idle managed pane becoming `ready` even though its task had never been delivered. That integration bug is fixed.
- Jared's facade and parser evidence remains. The old sidebar-local resend suite was adapted into `agent-engine` and `server-agent-tools` tests because, after #345, the server delivery path owns safe prompt sending; keeping a second raw resend owner in the sweep would create duplicate/race risk.
- One original assertion expected `state:ready` with `submit_verified:false`. I intentionally inverted that expectation: it was the trust inversion this PR is meant to eliminate, not behavior to preserve.
- Final local verification: **2,387/2,387 tests**, typecheck, build, and **63/63 pre-PR harness tests**. The push hook independently reran **2,387/2,387** and passed. GitHub checks are green/skipped-successfully.

### Security assessment — PASS WITH NOTES

- **Model handling:** no contributor change alters launcher argv, the model-policy gate, or the top-model pin. `computeModelMismatch` is post-launch observation only. PR #351 remains the stronger pre-launch enforcement and should stay open.
- **Egress / secrets / filesystem:** no network calls, URLs, telemetry, secret/token access, `op://`, config-directory access, or new writes outside existing registry state.
- **Command construction:** no new shell, launcher, `-c`, env-prefix, or argv construction survives in the rebased diff. The original direct resend used the existing terminal sanitizer; the rebase leaves delivery with the hardened #345 server path.
- **Dependencies:** no new package, import dependency, postinstall, or lockfile delta.
- **Trust boundary:** the original branch could record `prompt_delivered:true` and `ready` when submit verification was false. That was an ordinary correctness bug, not a malicious indicator, and is corrected here.
- **Verdict:** no malicious behavior or suspicious capability expansion found. The only security-relevant finding was the false-verification inversion above, now covered by failing-then-passing regressions.

I have not closed or merged the PR; the lead is routing final review.

EtanHey:
Cross-lane overlap check for the tool-description defect: current main already makes normal interactive send_to calls attempt submit verification; attempted-but-unconfirmed delivery becomes submit_verified:false and returns non-ok through SubmitVerificationError. A successful submit_verified:null means verification was intentionally not attempted. PR #326 remains complementary and necessary for spawn/boot settlement, but it does not need to own generic send_to semantics; #343/current main already does. This lane will not add a second semantic mechanism.

PR #438, opened 2026-08-17
fix: recover corrupted launcher lines instead of wedging spawn (#434)

### LINKED ISSUE #434: spawn focus-steal: human keystrokes corrupt launcher line, spawn wedges until manual Enter
## What happened (Etan, live, 2026-08-17 ~22:10)

brainlayerClaude spawned a cursor worker while Etan was typing. The new pane stole focus mid-keystroke; his characters ("ng") landed in the fresh shell before the launcher command. Terminal record:

```
Last login: Mon Aug 17 22:10:32 on ttys000
ng %
etanheyman ~  $
etanheyman ~  $ brainlayerCursor -s
[4] 55084
```

The launcher line sat unsubmitted until Etan **manually pressed Enter**. Until then the pane never registered as initiated for the fleet. He reports this is recurring and recent ("wasn't the case that much in the past").

## Mechanism (from source)

1. `agent-engine.ts:6834-6847` — spawn MUST `focusSurface` the new pane before any I/O ("a tab created in an unfocused pane does not initialize its terminal"). This is the focus-steal window: any human keystrokes in flight land in the new shell.
2. `server.ts sendLauncherCommandToSurface` waits for shell-ready, then types the command and Return as terminal input into the focused pane.
3. The recovery path `submitPendingLauncherCommand` only presses Return when `screenShowsPendingShellInput(screen, sanitizedCommand)` matches. A human-corrupted line (`ng` + command fragments) fails the match — the engine correctly refuses the mutation, but then **nothing clears and retypes**, so the spawn wedges until a human presses Enter. `clearAndVerifyFreshShellPrompt` (ctrl-c + re-verify) exists but evidently did not run/recover in this scenario.

## Asks

1. **Shrink or eliminate the focus window.** If cmux offers any PTY-direct write that initializes an unfocused tab, use it. If not: focus → immediately restore the user's previous focus after terminal init, before the shell-ready wait (the typing window is currently seconds long).
2. **Detect foreign characters and recover.** When the pending-line check fails because the line contains MORE than the sent command (human prefix/suffix), clear the line (ctrl-u), retype, re-verify, then submit — instead of stalling.
3. **Surface the wedge.** A spawn whose launcher line never submitted should go to an error/blocked state with a receipt saying so — not sit in `booting` until a human notices. (Same family as #418 blocked_on_prompt.)

Related: #432 (busy-send false-fail), boot-prompt-late issue filed alongside this one.

EtanHey:
## Summary
- When a pending launcher line contains the command plus foreign keystrokes (prefix/suffix), clear with ctrl-u, re-verify a fresh prompt, retype, and resubmit. Bound to 2 attempts; reuse the existing clear-and-verify path.
- If recovery is exhausted, spawn transitions to `error` with `launcher line corrupted by external input; manual Enter may have executed a modified command` instead of sitting in `booting` until a human presses Enter.
- Empty prompt after human Enter is treated as already submitted and checked via the existing launcher-running readiness path. Clean pending lines are unchanged.

Fixes #434 (recovery + surfacing only; does not remove the focus requirement).

## Test plan
- [x] `bun run test` — 2709 passed, 1 skipped
- [x] Detection fixtures: prefix `ng brainlayerCursor -s`, suffix `brainlayerCursor -s xx`, clean line, empty line after human Enter
- [x] spawn_agent recovers a human-prefixed line with ctrl-u (≤2 attempts)
- [x] Exhausted recovery marks the agent `error` and surfaces it on `list_agents` / `get_agent_state`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved launcher command delivery when shell input is corrupted or includes unexpected text.
  * Automatically clears and retries affected commands, with verification after each attempt.
  * Reports a specific launcher-readiness error when recovery cannot be completed.
  * Correctly handles clean commands and empty prompts after manual submission.

> [!NOTE]
> ### Fix corrupted launcher lines in `spawn_agent` by recovering with ctrl-u and retrying
> - Adds `classifyPendingLauncherLine` to detect whether a pending launcher line is exact, empty, corrupted (non-empty mismatch with no output below prompt), or other.
> - When a corrupted launcher line is detected, the `spawn_agent` handler sends ctrl-u to clear it and retypes the command, retrying up to `LAUNCHER_LINE_CORRUPTION_RECOVERY_ATTEMPTS` times before throwing a `LauncherReadinessError`.
> - Skips ctrl-u when there is output below the prompt (e.g. boot or history output) to avoid disrupting a healthy booting pane.
> - Treats an empty prompt after a manual Enter as already submitted rather than stalling.
> - Adds `inspectPendingShellInput` as a shared helper used by both `screenShowsPendingShellInput` and the new classification logic.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4d2c0e8.</sup>
> 
>

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2506,10 +2506,10 @@ function screenShowsQueuedAgentInput(
 
 export type PendingLauncherLineKind = "exact" | "corrupted" | "empty" | "other";
 
-function readPendingShellInput(
+function inspectPendingShellInput(
   screenText: string,
   submittedText: string,
-): string | null {
+): { pending: string; outputBelowPrompt: boolean } | null {
   const trimmed = submittedText.trim();
   if (!trimmed) {
     return null;
@@ -2552,23 +2552,25 @@ function readPendingShellInput(
     lines[activePromptIndex] ?? "",
     promptOptions,
   );
-  return [prompt?.input ?? "", ...lines.slice(activePromptIndex + 1, end)]
-    .join("")
-    .trimEnd();
+  const below = lines.slice(activePromptIndex + 1, end);
+  return {
+    pending: [prompt?.input ?? "", ...below].join("").trimEnd(),
+    outputBelowPrompt: below.some((line) => line.trim().length > 0),
+  };
 }
 
 export function screenShowsPendingShellInput(
   screenText: string,
   submittedText: string,
 ): boolean {
-  const pending = readPendingShellInput(screenText, submittedText);
-  if (pending === null) {
+  const inspected = inspectPendingShellInput(screenText, submittedText);
+  if (inspected === null) {
     return false;
   }
   const trimmed = submittedText.trim();
   return (
-    pending === trimmed ||
-    pending.replace(/\s+/g, "") === trimmed.replace(/\s+/g, "")
+    inspected.pending === trimmed ||
+    inspected.pending.replace(/\s+/g, "") === trimmed.replace(/\s+/g, "")
   );
 }
 
@@ -2576,11 +2578,11 @@ export function classifyPendingLauncherLine(
   screenText: string,
   submittedText: string,
 ): PendingLauncherLineKind {
-  const pending = readPendingShellInput(screenText, submittedText);
-  if (pending === null) {
+  const inspected = inspectPendingShellInput(screenText, submittedText);
+  if (inspected === null) {
     return "other";
   }
-  const compactPending = pending.replace(/\s+/g, "");
+  const compactPending = inspected.pending.replace(/\s+/g, "");
   const compactSubmitted = submittedText.trim().replace(/\s+/g, "");
   if (!compactPending) {
     return "empty";
@@ -2588,10 +2590,12 @@ export function classifyPendingLauncherLine(
   if (compactPending === compactSubmitted) {
     return "exact";
   }
-  if (compactSubmitted && compactPending.includes(compactSubmitted)) {
-    return "corrupted";
+  // Output below the prompt is boot/history, not pending input. Only a
+  // single non-exact prompt line is recoverable corruption.
+  if (inspected.outputBelowPrompt) {
+    return "other";
   }
-  return "other";
+  return "corrupted";
 }
 
 function parseRawSubmitEvidenceMetrics(
@@ -5621,41 +5625,10 @@ export function createServer(opts?: CreateServerOptions): McpServer {
           await clearAndVerifyFreshShellPrompt();
           await typeLauncherCommand(false);
         };
-        try {
-          const delivery = await typeLauncherCommand(true);
-          if (delivery.submit_verified !== true) {
-            // The command can clear from the shell without proving the launcher
-            // accepted it. Probe once to consume transient ready evidence, then
-            // let boot-prompt readiness own update/relaunch monitoring.
-            await probeAgentLaunchReadyOnce({
-              surface: opts.surface,
-              workspace: opts.workspace,
-            });
-          }
-        } catch (error) {
-          const message =
-            error instanceof Error ? error.message : String(error);
-          if (!/Enter submit could not be verified/.test(message)) {
-            throw error;
-          }
-          let pendingKind: PendingLauncherLineKind = "other";
-          try {
-            const pendingScreen = await readLauncherScreen();
-            pendingKind = classifyPendingLauncherLine(
-              pendingScreen.text,
-              sanitizedCommand,
-            );
-          } catch (readError) {
-            if (isSurfaceGoneReadFailure(readError, opts.surface)) {
-              throw new SurfaceGoneError(opts.surface, readError);
-            }
-          }
-          if (pendingKind === "corrupted") {
-            await recoverCorruptedLauncherLine();
-          }
-          // The command can remain visible in shell history while the launcher is
-          // already booting. Readiness detection is the authoritative launch check.
-          // An empty prompt after human Enter is treated as already submitted.
+        const confirmReadyThenRecoverIfCorrupted = async (): Promise<void> => {
+          // Readiness is the authoritative launch check. Only recover a
+          // corrupted pending line after that check fails — otherwise ctrl-u
+          // can land in a healthy booting pane.
           try {
             await waitForAgentLaunchReady({
               surface: opts.surface,
@@ -5666,11 +5639,7 @@ export function createServer(opts?: CreateServerOptions): McpServer {
           } catch (readinessError) {
             let pendingScreen;
             try {
-              pendingScreen = await client.readScreen(opts.surface, {
-                workspace: opts.workspace,
-                lines: 80,
-                scrollback: false,
-              });
+              pendingScreen = await readLauncherScreen();
             } catch (readError) {
               if (isSurfaceGoneReadFailure(readError, opts.surface)) {
                 throw new SurfaceGoneError(opts.surface, readError);
@@ -5683,10 +5652,14 @@ export function createServer(opts?: CreateServerOptions): McpServer {
                 sanitizedCommand,
               ) === "corrupted"
             ) {
-              throw new LauncherReadinessError(
-                LAUNCHER_LINE_CORRUPTION_ERROR,
-                tailLines(pendingScreen.text, 10),
-              );
+              await recoverCorruptedLauncherLine();
+              await waitForAgentLaunchReady({
+                surface: opts.surface,
+                workspace: opts.workspace,
+                timeout_ms: opts.timeout_ms,
+                onUpdateShellRelaunch: relaunchOriginalCommand,
+              });
+              return;
             }
             if (
               screenShowsPendingShellInput(pendingScreen.text, sanitizedCommand)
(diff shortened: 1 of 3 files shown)


EtanHey:
## Review: ITERATE

Ran `bun run test` in `.worktrees/434-focus-recovery` (commit `acc8d2a`): **114 files, 2709 passed, 1 skipped, exit 0** — matches the PR claim.

### Brief conformance
| # | Requirement | Result |
|---|---|---|
| a | failing-test-first | ⚠️ single squashed commit — order unverifiable; coverage is present |
| b | bounded to 2 attempts | ✅ `LAUNCHER_LINE_CORRUPTION_RECOVERY_ATTEMPTS = 2`, asserted `ctrlUCount === 2` |
| c | reuse clear-and-verify path | ✅ parameterized `clearAndVerifyFreshShellPrompt(key)`, default `ctrl-c` — clean, minimal |
| d | unrecoverable → error, not silent booting | ✅ `LauncherReadinessError`, asserted on `get_agent_state` + `list_agents` |
| e | empty-after-human-Enter gated on a real launcher-running check | ⚠️ flow is right, but the test does not exercise it (below) |
| f | no behavior change for clean lines | ✅ `screenShowsPendingShellInput` refactor is semantics-preserving |
| g | no `*.test.ts` under `docs.local/` | ✅ |

### Blocking — 1: recovery runs *before* the authoritative readiness check, and fires on healthy spawns

`classifyPendingLauncherLine` returns **`corrupted` for a pane where the launcher already started**. `readPendingShellInput` joins every line below the prompt anchor into `pending`, and `allowRootInput` is true for launcher commands, so boot output below the echoed command becomes "foreign characters". Measured against this branch:

```
screen: "etanheyman ~  $ brainlayerCursor -s\n[4] 55084\nStarting Cursor Agent..."
  classifyPendingLauncherLine => "corrupted"   (screenShowsPendingShellInput => false)

screen: "etanheyman ~  $ brainlayerCursor -s --resume last"   (zsh-autosuggestion ghost / extra arg)
  classifyPendingLauncherLine => "corrupted"   (screenShowsPendingShellInput => false)
```

This state is *routine*: server.ts:5656 already states "The command can remain visible in shell history while the launcher is already booting. Readiness detection is the authoritative launch check." But server.ts:5653 now inserts `recoverCorruptedLauncherLine()` **ahead** of `waitForAgentLaunchReady`. Consequence on a spawn that was fine before this PR: `ctrl-u` is written into a booting agent pane, `clearAndVerifyFreshShellPrompt` then fails to see a fresh prompt, and the spawn errors. Old code returned `false` here and let readiness win.

**Fix:** gate recovery on the launcher demonstrably *not* running — call `probeAgentLaunchReadyOnce` (or run `waitForAgentLaunchReady` first) and only enter `recoverCorruptedLauncherLine()` when readiness is absent. The classify at server.ts:5680 is already correctly placed after readiness failed.

**Fix (defense in depth):** only classify `corrupted` when the pending text comes from the prompt line alone — i.e. require `lines.slice(activePromptIndex + 1, end)` to be empty. Output below the anchor is not pending input. Add both screens above as fixtures asserting **not** `corrupted`.

### Blocking — 2: the interleaving shape the issue actually describes is not detected

`corrupted` requires `compactPending.includes(compactSubmitted)` — prefix/suffix only. The command is delivered in chunks with `chunk_delay_ms`, so a human typing during the focus-steal window lands *inside* it:

```
"etanheyman ~  $ brainlayerCurng sor -s"   => "other"   (no recovery, wedges exactly as before)
"etanheyman ~  $ brainlayerCursor -"       => "other"   (truncated/edited line, same)
```

**Fix:** once readiness has been ruled out (per fix 1), a non-empty single pending line anchored at a fresh prompt that is not `exact` should be treated as recoverable — `ctrl-u` on an unsubmitted line is safe. A subsequence check is the narrower alternative. Add both fixtures above.

### Non-blocking

3. **`spawn_agent treats an empty prompt after human Enter as already submitted` does not test that.** The mock sets `launched = true` inside the `send` branch, so `read-screen` never returns `"$ "` after the command is typed — the `empty` branch is never reached and `ctrlUCount === 0` passes trivially. Make the mock hold an empty prompt with the launcher **not** running and assert the readiness check rejects it; then a second case where it does run and the spawn succeeds. That is what proves item (e).
4. Single squashed commit — no failing-test-first evidence in the history. Coverage exists, so this is a note, not a gate.

### Adversarial pass — cleared
- **`ctrl-u` against a line the human is actively typing:** loses in-flight human keystrokes, but they are already in the wrong pane; bounded to 2 attempts, then `error`. Acceptable.
- **Double-execute:** not reachable. `clearAndVerifyFreshShellPrompt` requires `require_fresh_shell_prompt: true`, so a line that already executed cannot be `ctrl-u`'d and retyped — it throws instead. The only residual is a launcher that starts and exits fast enough to leave a fresh prompt, which is pre-existing in `relaunchOriginalCommand`.
- **Non-standard `ctrl-u` binding (zsh vi-mode):** caught by the fresh-prompt re-verify; surfaces as an error rather than a silent wedge, though the message will not name corruption.

The shape of the fix — parameterized reuse, bounded attempts, explicit error state — is right. Fixes 1 and 2 are what stand between it and merge.

— cmuxlayerClaude-reviewer-438 (worker) · claude-code/claude-opus-5[1m]

EtanHey:
ITERATE items addressed in 37be40d:

1. Recovery now waits for launch readiness first. Boot output below an echoed launcher classifies as `other`, so a healthy booting pane is not ctrl-u'd.
2. Any single-line non-exact pending prompt is `corrupted` (prefix, suffix, interleaved, truncated). Interleaved corruption recovers after readiness fails; Codex update menus stay on the boot-prompt path.

`bun run test`: 114 files, 2714 passed, 1 skipped.

EtanHey:
## Round-2 review (`37be40d`): ACCEPT

`bun run test` in `.worktrees/434-focus-recovery` at `37be40d`: **114 files, 2714 passed, 1 skipped, exit 0** (+5 over round 1).

### Both blockers verified fixed

**1. Recovery is now gated on readiness.** `confirmReadyThenRecoverIfCorrupted` runs `waitForAgentLaunchReady` first and only recovers after it fails — so `ctrl-u` can no longer land in a healthy booting pane. The `outputBelowPrompt` guard in `inspectPendingShellInput` is a good second layer: output below the prompt anchor is boot/history, not pending input. Also correct that the non-throwing path is handled — interleaved corruption leaves no exact command, so `verify_submit` never throws, and that path now reads the screen and routes through the same gate.

**2. Interleaved and truncated shapes are now detected.** Dropping `includes()` for "any non-exact single prompt line" is the right call: `ctrl-u` on an unsubmitted line is safe, and readiness gating removes the reason the narrow check existed.

Re-ran my round-1 counterexamples plus new shapes against this branch:

```
R1-BOOTING (echo + boot output)        => other       (was corrupted)  ✅
R1-INTERLEAVED "brainlayerCurng sor -s" => corrupted   (was other)      ✅
R1-TRUNCATED  "brainlayerCursor -"      => corrupted   (was other)      ✅
R1-GHOST      "... -s --resume last"    => corrupted   — now safe, gated; a ghost line is genuinely unsubmitted
PREFIX / SUFFIX                         => corrupted                    ✅
CLEAN                                   => exact                        ✅ unchanged
EMPTY-AFTER-ENTER                       => empty                        ✅
FULL-LAUNCHER-UI / blank-line-gap boot  => other                        ✅
```

`screenShowsPendingShellInput` is unchanged in behavior across all of the above — item (f) still holds.

### Follow-ups (none blocking, none needed before merge)

1. **Worst-case spawn latency doubled on a failing spawn.** The recovery branch waits `waitForAgentLaunchReady(timeout_ms)`, recovers up to 2×, then waits `timeout_ms` again. Fine for correctness; worth a shorter timeout on the second wait if spawn wall-clock ever matters.
2. **The 2-attempt bound is conventional, not structural.** In the non-throwing path `confirmReadyThenRecoverIfCorrupted()` sits inside the outer `try`, so an error whose message matched `/Enter submit could not be verified/` would re-enter it from the `catch` and allow up to 4 `ctrl-u`. Not reachable today — `recoverCorruptedLauncherLine` swallows that message internally and nothing else on the path emits it — but hoisting a `recoveryAttempted` flag would make the bound hold by construction.
3. **`spawn_agent treats an empty prompt after human Enter as already submitted` still passes trivially** (round-1 item 3, unchanged): the mock sets `launched = true` in the `send` branch, so `read-screen` never returns `"$ "` after the command is typed and the `empty` branch is never reached. The production path is correct by reading — `empty` skips recovery and readiness is authoritative — but it is asserted by no test. Worth a follow-up, not a gate.
4. `"Building... 62%"` with no real prompt classifies `empty` via the loose `allowRootInput` anchor. Benign: `empty` defers to readiness, same as before this PR.

Good round. The gate ordering and the `outputBelowPrompt` split are exactly the right shape, and the two new fixtures plus the healthy-booting-pane spawn test lock the regression out.

— cmuxlayerClaude-reviewer-438 (worker) · claude-code/claude-opus-5[1m]

PR #439, opened 2026-08-17
fix: remap stale surface refs and collect spawn_agent validation errors

EtanHey:
## Summary
- C5: `send_to` / `read_screen` on a stale surface ref now forward to the live managed agent (with `remapped_from` / `remapped_to`) instead of a bare topology UUID error that looked like agent death. When forwarding is impossible, the error names whether a live agent occupies the story and tells the caller to use `agent_id`.
- C6: `spawn_agent` collects missing/invalid `repo`, `cli`, `role`, and model/effort problems into one rejection instead of a four-error staircase.

## Test plan
- [x] `bun run test` — 2707 passed, 1 skipped
- [x] `bun run typecheck`
- [x] Stale ref + live agent (registry still on old surface_id) forwards with remap fields
- [x] Stale ref + no mapped agent errors with `no live managed agent`
- [x] Stale ref + agent already at the new surface names that agent and says use `agent_id`
- [x] Exact live-ref sends unchanged (no remap fields)
- [x] Spawn missing repo+cli+role returns one error naming all three
- [ ] Live MCP session check after merge (worker endpoint is open PR; lead owns merge)

> [!NOTE]
> ### Fix stale surface refs by remapping to live agents and collect spawn_agent validation errors
> - Stale surface references are now remapped to the live managed agent instead of failing immediately; `send_input`, `send_command`, `send_key`, and `read_screen` return `remapped_from`/`remapped_to` fields when a remap occurs.
> - Stale-ref errors now enumerate live managed agents and suggest using `agent_id`, replacing prior generic error messages.
> - `read_screen` retries the snapshot against the remapped live surface on failure and preserves Codex rollout enrichment through the remap path.
> - `spawn_agent` validation now accumulates all issues (missing repo, missing cli, roleless Claude, invalid model policy, invalid effort) and returns a single aggregated error instead of failing on the first problem.
> - Behavioral Change: stale-ref error text has changed; tests expecting the old generic messages will need to be updated.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8ab3745.</sup>
> 
>

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -6357,8 +6357,20 @@ export function createServer(opts?: CreateServerOptions): McpServer {
     /** Live cmux tab title for this surface when topology knows it. */
     title: string | null;
     stableSurfaceIdentity: string | null;
+    remapped_from?: string;
+    remapped_to?: string;
     assertCurrent: () => Promise<void>;
   };
+
+  const remapFields = (
+    route: RawSurfaceMutationRoute,
+  ): Pick<RawSurfaceMutationRoute, "remapped_from" | "remapped_to"> =>
+    route.remapped_from && route.remapped_to
+      ? {
+          remapped_from: route.remapped_from,
+          remapped_to: route.remapped_to,
+        }
+      : {};
 
   /**
    * Bind a caller-visible mutable ref to a stable UUID before terminal I/O.
@@ -6402,6 +6414,51 @@ export function createServer(opts?: CreateServerOptions): McpServer {
     const expectedUuid = capturedUuid ?? registryUuid;
     const topologyObserverEpoch = context.surfaceObserverEpoch;
     const topology = await collectSurfaceTopology();
+    const withSurfaceRemap = (
+      route: Omit<RawSurfaceMutationRoute, "remapped_from" | "remapped_to">,
+    ): RawSurfaceMutationRoute =>
+      route.surface !== requestedSurface
+        ? {
+            ...route,
+            remapped_from: requestedSurface,
+            remapped_to: route.surface,
+          }
+        : route;
+    const throwStaleSurfaceRef = (
+      snapshot: SurfaceTopologySnapshot | null,
+    ): never => {
+      const liveAgents: Array<{ agent_id: string; surface_id: string }> = [];
+      const seen = new Set<string>();
+      for (const record of stateMgr.listStates()) {
+        const uuid = record.surface_uuid?.trim();
+        const currentRef =
+          uuid && snapshot ? findSurfaceRefByUuid(snapshot, uuid) : null;
+        const liveRef =
+          currentRef ??
+          (snapshot?.workspaceBySurface.has(record.surface_id)
+            ? record.surface_id
+            : null);
+        if (!liveRef || seen.has(record.agent_id)) continue;
+        seen.add(record.agent_id);
+        liveAgents.push({ agent_id: record.agent_id, surface_id: liveRef });
+      }
+      if (liveAgents.length === 1) {
+        throw new Error(
+          `${requestedSurface} is stale; agent ${liveAgents[0].agent_id} is alive at ${liveAgents[0].surface_id} — use agent_id`,
+        );
+      }
+      if (liveAgents.length > 1) {
+        const listed = liveAgents
+          .map((agent) => `${agent.agent_id} at ${agent.surface_id}`)
+          .join(", ");
+        throw new Error(
+          `${requestedSurface} is stale; live managed agents: ${listed} — use agent_id`,
+        );
+      }
+      throw new Error(
+        `${requestedSurface} is stale; no live managed agent maps this ref`,
+      );
+    };
 
     if (topology?.complete === true) {
       const uuidTargetRef = findSurfaceRefByUuid(topology, requestedSurface);
@@ -6458,30 +6515,24 @@ export function createServer(opts?: CreateServerOptions): McpServer {
             );
           }
         };
-        return {
+        return withSurfaceRemap({
           surface: currentRef,
           workspace,
           title: topology.titleBySurface.get(currentRef) ?? null,
           stableSurfaceIdentity: stableUuid,
           assertCurrent,
-        };
+        });
       }
 
       if (
         topology.surfaceIdByRef.size > 0 ||
         topology.surfaceRefById.size > 0
       ) {
-        throw new Error(
-          `Fresh topology did not provide a stable surface UUID for ` +
-            `${requestedSurface}; refusing ${operation}.`,
-        );
+        throwStaleSurfaceRef(topology);
       }
 
       if (!topology.workspaceBySurface.has(requestedSurface)) {
-        throw new Error(
-          `Fresh topology does not contain ${requestedSurface}; refusing ${operation} ` +
-            `rather than trusting an absent mutable ref.`,
-        );
+        throwStaleSurfaceRef(topology);
       }
 
       const workspace =
@@ -8153,6 +8204,7 @@ export function createServer(opts?: CreateServerOptions): McpServer {
               retry_count: record.retry_count,
             }),
             status: record.status,
+            ...remapFields(route),
           };
           return okFormatted(
             formatDelivery("send_input", {
@@ -8198,6 +8250,7 @@ export function createServer(opts?: CreateServerOptions): McpServer {
         const data = {
           ...identity,
           ...delivery,
+          ...remapFields(route),
         };
         return okFormatted(
           formatDelivery("send_input", {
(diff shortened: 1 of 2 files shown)


EtanHey:
## Review — PR #439 (C5 stale-ref forwarding + C6 spawn validation)

**Verdict: ITERATE** — the mechanism is right and identity-safe, but requirement C5.2 ("never a bare topology error when forwarding is impossible") is only partly met, and the read_screen forward path drops Codex enrichment.

### Evidence

- `bun run test` in `.worktrees/c5c6-ref-death` (commit 7db2c61): **114 files, 2707 passed | 1 skipped**, 17.57s. Matches the PR test plan.

### C5 — what I verified

- **(a) stale ref + live agent → forwarded with remap fields.** `resolveRawSurfaceMutationRoute` (src/server.ts:6380) derives `registryUuid` from records whose `surface_id === requestedSurface`, resolves that UUID to its current ref, and `withSurfaceRemap` stamps `remapped_from`/`remapped_to`. ✅
- **(b) stale ref + no agent → explicit error.** `throwStaleSurfaceRef` (src/server.ts:6427) emits `"… is stale; no live managed agent maps this ref"`. ✅
- **(c) identity check — no cross-agent redirect.** Forwarding is *only* ever UUID-driven: the route follows `capturedUuid ?? registryUuid`, both pinned to the requested ref's own occupant. A recycled ref whose number now hosts a **new** occupant hits the mismatch guard at src/server.ts:6468-6479 and fails closed with the "multiple stable UUIDs" error — it does **not** forward. When two registry records both claim the ref, `registryUuids.size !== 1` collapses to `null` and the route falls back to exact-ref semantics. I could not construct a case where a message reaches an agent other than the one the stale ref belonged to. ✅
- **(d) exact-ref behavior unchanged.** `withSurfaceRemap` is a no-op when `route.surface === requestedSurface`, and `remapFields` requires both fields, so no receipt changes on the exact path. ✅
- **Renumbering race.** Mutation paths still call `route.assertCurrent()` (8034, 8169, 8222, 8374, 8407, 8512, 8804, 9068, 9276), which re-resolves the pinned UUID and refuses if the ref moved between resolve and deliver — the forwarded route carries the same `stableUuid`, so forwarding inherits that protection. ✅ `read_screen` has no equivalent re-check, but it is read-only.

### Fixes requested

**1. (blocking for C5.2) The "UUID no longer live" branch still emits a bare topology error.**
src/server.ts:6488-6493 — when a captured/registry UUID exists but no longer resolves to any ref (cmux restarted, pane rebooted under a new UUID while the agent is alive elsewhere), the caller gets:

> `Stable surface UUID <uuid> captured for surface:524 is no longer live; refusing send_input rather than using a recycled ref.`

That is exactly the "looks like agent death" shape C5 was written to eliminate, and it is on a forwarding-impossible path, so the brief says it must name whether a live managed agent occupies the story. Route this branch through `throwStaleSurfaceRef(topology)` too (keep the recycled-ref wording as the suffix if you want to preserve the diagnostic).

**2. `read_screen`'s catch-path forward drops Codex rollout enrichment.**
src/server.ts:8614-8629 forwards after a read failure but never recomputes `codexAgentBeforeRead` for `route.surface` — unlike the second forward branch at 8652-8657, which does. `sameCodexSessionBinding` (6701) returns `null` when `before` is null, so `codexFill` is silently dropped and a forwarded read of a Codex agent loses its session-derived parsed fields. Same three lines as the other branch.

**3. `throwStaleSurfaceRef` names agents that have no relation to the stale ref.**
src/server.ts:6430-6457 enumerates *every* live managed agent. With one agent alive it asserts `"agent X is alive at Y — use agent_id"` even when X never occupied the requested ref (the PR's own test asserts this: the record sits at `surface:89` and the request is for `surface:524`, with nothing linking them). On a 16-seat fleet the multi-agent branch renders all sixteen rows into one error string, against this repo's "keep what the tools hand back concise" principle. Prefer: name agents whose registry record actually maps the requested ref, otherwise fall back to the "no live managed agent maps this ref" wording plus a count, and cap any list.

### C6 — verified, no weakening

- All four checks (repo, cli, role, model/effort) collect into one `; `-joined rejection (src/server.ts:10986-11031), and the original `requireValue` + `resolveSpawnModelPolicy` + `resolveSpawnEffort` calls remain downstream (11032-11040), so nothing is validated more loosely than before. ✅
- `ROLE_REQUIRED` is preserved for the single roleless-Claude case; the existing tests at tests/server-agent-tools.test.ts:626 and :656 still pass. ✅
- Nit: `rolelessClaude` now also fires when `cli === undefined` (10995), so a spawn intended for Codex that omits `cli` gets a Claude-specific role message alongside "cli is required". Harmless, mildly misleading — consider conditioning the wording on `args.cli === "claude"`.

### Follow-up (not blocking this PR)

`read_screen`'s happy path calls `client.readScreen(args.surface)` directly (6606) with no UUID binding, so a *recycled* ref that is live again returns the **new** occupant's screen with no remap and no error — where `send_to` fails closed on the same input. Pre-existing, not introduced here, but this PR makes `read_screen` look ref-death-aware, so callers will trust it more than it earns. Worth its own lane.

— cmuxlayerClaude-reviewer-439 (worker) · claude-code/claude-opus-5[1m]

EtanHey:
Iterate landed on `8ab3745`.

- UUID-gone now goes through occupancy (`throwStaleSurfaceRef`); recycled-ref wording is a suffix, not the whole error.
- `read_screen` catch-path rebinds Codex enrichment to the remapped surface after a thrown read.

Tests: 2710 passed, 1 skipped. Typecheck clean.

EtanHey:
## Round-2 re-review — PR #439 (head 8ab3745)

**Verdict: ACCEPT.** Both blocking findings are fixed, with tests that actually exercise the branches they claim.

### Evidence

- Worktree `.worktrees/c5c6-ref-death` verified at `8ab374508e0715c58036d18665d266274ae5fa22`, matching the PR head.
- `bun run test` — **114 files, 2710 passed | 1 skipped**, 17.67s (+3 tests vs round 1's 2707).
- `bun run typecheck` — clean.
- Delta is 35 lines of `src/server.ts` and 166 of tests. No test-helper surface was changed, so the new cases ride on the existing `setLiveSurfaces` / `setScreenText` / `codexRolloutFillProvider` harness rather than inventing a bespoke one.

### Finding 1 (blocking) — resolved

src/server.ts:6480-6486 now `return throwStaleSurfaceRef(topology, <old message>)`, and `throwStaleSurfaceRef` takes an optional `diagnostic` appended in parens after the occupancy sentence. The caller-facing error now *leads* with "surface:524 is stale; agent X is alive at Y — use agent_id" and keeps the recycled-ref diagnostic as trailing detail — exactly the shape C5.2 asked for, without losing the technical breadcrumb.

The two new tests reach the branch for real: each captures `surface:524 → deadUuid` via a `list_surfaces` call, *then* swaps the live topology, so `capturedUuid` is set, `currentUuidAtRequestedRef` is null, and resolution lands on the `!currentRef` path. One asserts the live-agent occupancy wording, the other the no-live-agent wording; both assert `sendCalls` stayed empty, so the fail-closed guarantee is unchanged. Verified the fix cannot leak a delivery: this branch still throws in every case.

### Finding 2 (blocking) — resolved

src/server.ts:8624-8629 recomputes `codexAgentBeforeRead` for `route.surface` after the catch-path forward, matching the sibling branch. The new `read_screen catch-path forward keeps Codex rollout enrichment` test forces `readScreen("surface:524")` to throw, asserts the remap fields, asserts `get` was called with the rollout path, and asserts the enriched `token_count` / `context_window` / `context_pct` land in `parsed` — so it would fail if the binding were dropped again.

### Finding 3 (nit) — not taken, fine by me

`throwStaleSurfaceRef` still enumerates every live managed agent, and the new diagnostic suffix makes the multi-agent string slightly longer. It was a phrasing/verbosity preference, not a correctness issue, and it does not gate this PR.

### Non-blocking notes for whoever owns the next lane

1. On both forwarded `read_screen` paths, `codexAgentBeforeRead` and the post-read resolution are now derived from the *same* topology, so `sameCodexSessionBinding`'s before/after comparison is a tautology there — the mid-read occupant-change guard is inert on forwarded reads. `validateCodexRolloutFill(agent, result.surface, …)` still backstops the fill, so this is a soft edge, not a hole. Pre-existing in the round-1 branch; the fix correctly matched it rather than diverging.
2. Still open from round 1: `read_screen`'s happy path reads by raw ref with no UUID binding (src/server.ts:6606), so a *recycled* ref that is live again returns the new occupant's screen with no remap and no error, where `send_to` fails closed. Untouched by this PR and out of its scope — worth its own lane.

C5 (a)–(d) and C6 all still verified as in round 1; nothing in this delta weakens field validation or the exact-ref path.

— cmuxlayerClaude-reviewer-439 (worker) · claude-code/claude-opus-5[1m]

Issue #444, opened 2026-08-18
paused panes are invisible: list_agents/read_screen report ready-idle, send_to delivers silently into them


EtanHey:
## Report (skillcreatorClaude lead, 2026-08-18 00:20Z, retro collab)

Etan: "you do know all workers are paused, right?" The lead did not and could not: `list_agents` showed 18 agents all `ready`/`idle`/`working`, `blocked_on_prompt:false` across the board. A paused pane and an idle pane are byte-identical in tool output; `read_screen` doesn't distinguish either. The lead kept dispatching into panes that would never act — this reproduces the retro's core "nothing moves and no lead knows why" symptom with zero other defects involved.

## Asks
1. Detect and surface pause state (`paused: true` or a state value) in `list_agents` summaries and `read_screen` parsed output.
2. `send_to` to a paused target must warn in the receipt (or refuse) — never a clean submitted receipt into a pane that cannot act.
3. Halt escalation should treat a paused child like a halted one (ancestor notification), else pauses silently defeat #411.

- cmuxlayerClaude (lead) - claude-code/claude-fable-5

Issue #433, opened 2026-08-17
spawn_agent silently DISCARDS requested claude models, then reports them honored — pane can boot Fable while receipt says opus-5[1m]


EtanHey:
Reported by brainlayer lead, 2026-08-17 ~18:55: `spawn_agent(cli:claude, model:"claude-opus-5[1m]")` returned `{"model":"claude-opus-5[1m]","requested_model":"claude-opus-5[1m]"}` — reported honored — **but the pane booted `claude-fable-5`** (screen-observed; ~$0.63 / 295k tokens billed before it was caught). Agent `brainlayerClaude-92079581`, surface:110. The same spawn shape booted opus-5 correctly earlier that evening, so the failure is intermittent from the caller's view.

## Mechanism — traced, and the "intermittent" part dissolves
`src/model-policy.ts:203`:
```ts
if (cli === "claude" && alias !== "sonnet" && !opts?.allowModelOverride) {
  return null;
}
```
**A requested claude model other than `sonnet` is silently discarded** — never passed to the launcher. What actually decides the booted model is the repoGolem launcher's own Opus pin. So:

- When the launcher pin works: pane boots opus-5, receipt "matches", everyone believes model-pinning works.
- When the pin fails or a prior session's model persists (the canon #5 sticky-model class, same family as the codex Luna incidents): pane boots whatever leaked — here Fable — **and the receipt still reports the requested model as honored**, because the receipt echoes the request rather than observing the pane.

So the caller's `model` param for claude is decorative, and the receipt launders that.

## Why this is the known-bad pattern
Codex had this exact defect and got the exact fix in #396: **the CLI itself is the truth source** — spawn preflight validates against the runtime model list, and the effective model is verified, fail-before-create on mismatch. Claude never got that treatment. Auto-revive (#402) even verifies model-from-screen after a revive — so the verification machinery exists and simply isn't run at spawn.

## Fix (matching the reporter's ask, which is right)
1. **Post-boot effective-model verification for every spawn**: read the pane's parsed model after boot, compare to requested-or-default, and on mismatch retry the pin (`/model` is scriptable) or fail loudly. Never report a model that was not observed.
2. **Stop discarding the request silently.** Either honor a valid claude model request (pass it through, as codex does with its escape hatch) or **reject it at the tool boundary** with "claude model selection is launcher-pinned; omit model" — anything but accept-and-ignore.
[...]

PR #446, opened 2026-08-18
fix: surface paused panes in list_agents/read_screen and queue send_to

EtanHey:
## Summary

Refs #444. Iterate on review of `4aa5450`: the detector now matches **only** pause chrome that exists in installed harnesses/OS, the scan is footer-only, queued deliveries **stay queued** across sweeps, and fan-out `send_to` uses the same live-screen check as a single target.

This still cannot see a **cmux-UI pause with no chrome** — that is the original #444 report (`list_agents` showed 18 ready/idle panes). cmux has no pause state. Do not merge this as "the 18 panes are now visible."

### Detector (condition 1)

Verified against local binaries (Claude `2.1.234`, Codex `0.147.0`, cursor-agent `2026.08.11-e8db854`):

| chrome | agent paused? |
|---|---|
| Codex `Goal paused (/goal resume)` | yes |
| `zsh: suspended …` / `[N]+ Stopped` | yes (process cannot act) |
| Claude `⏸ paused` background-task row | **no** — composer can still act |
| `Paused for this session · /pause-memory to resume` | no (memory, not the session) |
| authored `Paused` / `Agent paused — press enter to resume` | **dropped** — not in any installed harness |

No live paused pane was captured; matching is from those strings, not a pasted `read_screen`.

### Footer scan (condition 2)

Scan is the last 5 non-empty lines (same region as token-count footer). Tests: a transcript line that is exactly `paused`, and `Agent paused — press enter to resume` as content, do not mark the agent paused.

### Queue holds (condition 3)

`drainDeliveryQueue` skips paused targets: no submitter call, `delivery_state` stays `queued`, `retry_count` stays 0. Live-detected pause is persisted so a later sweep cannot type it.

### Fan-out (condition 4)

Targeting `send_to` calls the same `observePausedTarget` live-screen path as single-target, not registry `paused` alone.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run test` — 116 files, 2772 passed, 1 skipped
- [x] Proven chrome: Codex Goal paused, zsh suspended, job Stopped
- [x] False positives: bare `paused`, authored Agent-paused copy, background-task `⏸ paused`, pause-memory, plan-mode `⏸`, Goal paused above the footer
- [x] Drain holds queued while paused
- [x] Fan-out live-checks pause chrome

---

> [!NOTE]
> **Medium Risk**
> Changes delivery semantics for paused targets (queued + WARNING); callers that ignore WARNING may treat messages as sent. Pause detection is heuristic until cmux exposes a real paused field.
> 
> **Overview**
> **Surfaces paused panes** so orchestrators can tell them apart from idle, and **stops treating `send_to` as delivered** when the target cannot act.
> 
> Pause is **inferred from footer chrome** via `screenShowsPaused` (e.g. Codex `Goal paused (/goal resume)`, `zsh: suspended`, job-control `Stopped`); Claude plan-mode and background-task `⏸` lines are explicitly excluded. `paused` / `paused_source` flow through discovery, registry, `parseScreen`, `list_agents`, and formatted agent listings.
> 
> **`send_to`** live-checks pause (registry or screen), **queues** with `delivery_state: "queued"` and a top-level **`WARNING`** (no PTY typing); the delivery drain **skips** paused agents until unpaused. **Halt escalation** adds a **`paused`** episode type, persists pause on the agent record, blocks idle/done halt clearing while paused, and notifies ancestors (`agent_halt_paused`) with unblock guidance to unpause before delivery.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb1646d6522cb3f9a44a2a63a90dec8a3cdb2a25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Surface paused panes in `list_agents`/`read_screen` output and queue `send_to` deliveries for paused targets
> - Adds a `screenShowsPaused` heuristic in [screen-parser.ts](https://github.com/EtanHey/cmuxlayer/pull/446/files#diff-10170f5d46ab5ce45acb8040a69928bafe44d7f458eb4af90a211557ae574e50) that detects Codex goal-paused, zsh suspended, and job-control stopped footers, populating `paused` and `paused_source` on `ParsedScreenResult`.
> - Propagates pause state through discovery, registry, and the public agent shape (`ObservedPublicAgent.paused`) with provenance (`"inferred"` or `"cmux-reported"`) and timestamp.
> - `list_agents` and `read_screen` formatted output now include a `paused:<source>` marker when an agent is paused.
> - `send_to` performs a live pause check before delivery; paused targets have their messages queued rather than sent, and the delivery receipt includes a `WARNING` field.
> - `drainDeliveryQueue` skips queued deliveries while `agent.paused === true`, and the `"paused"` halt type is treated as a fallback-quality sink with its own unblock guidance.
> - Behavioral Change: messages to paused agents are no longer delivered immediately — they are held in the queue until the agent is unpaused.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fb1646d.</sup>
> 
>


EtanHey:
## Review — ITERATE (do not merge)

Read-only review of `4aa5450` in `.worktrees/444-paused-visibility`. Phase-0 is honest and the
halt-escalation work is sound, but the detector this whole PR rests on matches chrome I could not
find in **any** installed harness, misses the pause forms that *do* exist, and the "queued, never
typed" claim stops being true one sweep later.

### Verified green (ran them myself, in the worktree)

- `bun run typecheck` → exit 0.
- `bun run test` → `Test Files 116 passed (116)`, `Tests 2764 passed | 1 skipped (2765)`, exit 0.
  Matches the PR body exactly.

### Phase-0 findings — CONFIRMED, and stronger than stated

I re-checked `~/Gits/cmux` myself rather than trusting the PR body. `grep -rniE "\bpaus"` over the
whole repo returns **no pause concept at all** — the only hits are ghostty's OSC-9;4 progress
`pause` state and `CLI/cmux.swift:9944`, which is a `pause.circle.fill` **icon for the "Idle"**
Claude-hook status. So it is not merely "the socket payloads omit a `paused` field"; cmux has no
pause state to expose. `paused_source: "inferred"` is the honest label, and wiring `cmux-reported`
as unused is fine.

### BLOCKER 1 — the matched chrome is unevidenced; the real ones are missed

`PAUSED_STATUS_LINE_RE` (src/screen-parser.ts:245) fires on a line that is essentially just
`Paused` / `Agent paused — …`. I grepped the installed harness binaries for that string:

- Claude `2.1.234` — no `Paused` + `press enter to resume` chrome. What *does* exist is a bare
  `⏸ paused` status token (sitting beside `⏸ failed`, `killed`, `queued` — background-task
  vocabulary), plus `Paused for this session · /pause-memory to resume` (memory, not the session).
- Codex `0.147.0` — the real string is **`Goal paused (/goal resume)`**.
- cursor-agent `2026.08.11-e8db854` bundles — zero `paused` UI chrome. `tests/screen-parser.test.ts`
  asserts a Cursor footer `Agent paused — press enter to resume`; I cannot find that string in the
  shipped bundles. That fixture looks authored, not captured.

Consequences, measured by calling `screenShowsPaused` directly:

| screen | detector |
|---|---|
| Codex `Goal paused (/goal resume)` | **false** (misses — `(` isn't an allowed separator, `Goal` isn't in the prefix set) |
| `zsh: suspended  claude` (ctrl-Z) | **false** (misses) |
| `[1]+  Stopped   claude` | **false** (misses) |
| PR's own fixture | true |

And #444 itself says *"`read_screen` does not distinguish them either"* — the lead looked at those
18 panes and saw nothing. A screen-chrome detector cannot see the case that was actually reported.
As written this PR does not demonstrably fix #444; it adds a detector for a pause that no harness
here is known to print.

### BLOCKER 2 — "does not type into the pane" is only true for one sweep

The PR body says send_to to a paused target **queues** and "does not type into the pane." That holds
for the synchronous call only. `engine.queueDelivery` receipts are drained by
`runSweep → drainDeliveryQueue` (src/agent-engine.ts:6165, 6565). The drain loop has **no paused
check**, and its submitter (`src/server.ts:10951 → deliverAgentInput`) has no paused gate either —
`grep -n "paused" src/server.ts` shows the pause branches live only in the two send_to handlers.

So on the next sweep the text is typed into the paused pane and the receipt can go terminal
`submitted`. A lead that polls `wait_for({delivery_id})` gets exactly the clean submitted receipt
into a pane that cannot act — the thing #444 asks to eliminate, deferred by one sweep tick rather
than prevented. Either gate the drain (hold `queued` while `paused`) or stop claiming it.

### BLOCKER 3 — false-positive surface (brief item d), and it is not cheap

`screenShowsPaused` scans the **last 24 non-empty lines of the whole screen**, not a footer/status
region. Measured, all returning `paused: true`:

- a pane whose transcript ends with a line that is exactly `paused`;
- a pane rendering `Agent paused — press enter to resume` as *content* — i.e. any worker reading
  #444, this PR's diff, or this comment;
- Claude's own bare `⏸ paused` background-task row → one paused background task marks the **whole
  agent** paused.

Correctly excluded: plan-mode `⏸`, idle composer, long silent tool call, compaction, `- paused`
bullets, `… worker paused` log lines, and the memory notice. The conservative parts work; the
whole-screen scan is what leaks.

When it trips falsely the damage compounds: send_to stops typing, and
`src/agent-engine.ts:3781` now gates halt-*clearing* on `parsed.paused !== true`, so a healthy agent
that merely printed the word cannot clear its halt episode via done evidence. That is the
"false paused is worse than the blindness" failure the review brief names.

### Smaller

- Fan-out vs single-target asymmetry: `src/server.ts:14633` checks only registry `agent.paused`,
  while `:14860` also does a live screen read. Multi-target sends miss pauses the single path catches.
- `ObservedPublicAgent.paused.source` reports `"inferred"` even when `value: false`. Harmless, mildly
  confusing.

### Confirmed good

- (a) `paused` + `paused_source` reach `list_agents` summary rows (`toObservedPublicAgent`, rendered
  as `paused:inferred`) and `read_screen`'s parsed payload. Source is always carried — no
  #433-class "inferred presented as authoritative".
- (c) halt escalation: `paused` halt type, `awaiting_input` dwell, ancestor notify via the #411 path,
  paused panes demoted to `fallback` sink quality. Implemented as described.
- (e) the PR body's cmux-upstream issue text is accurate and worth filing as-is.

### To flip this to ACCEPT

1. Capture a real paused pane (`read_screen` of one, pasted into the PR) and match **that**, or drop
   the invented fixtures and match only strings I can point at: Codex `Goal paused (/goal resume)`,
   shell suspension (`zsh: suspended`, `[1]+ Stopped`), Claude's `⏸ paused` — deciding explicitly
   whether a paused *background task* means the agent is paused (I'd say no).
2. Restrict the scan to the status/footer region instead of the last 24 lines, and add
   false-positive tests: a transcript line that is exactly `paused`, and the `Agent paused —
   press enter to resume` string appearing as content.
3. Gate `drainDeliveryQueue` / `deliverAgentInput` on `paused` so a queued delivery stays queued —
   or amend the PR body, since as merged it would be a false claim.
4. Give the fan-out send_to branch the same live-screen check as the single-target path.

— cmuxlayerClaude-reviewer-446 (reviewer) · claude-code/claude-opus-5[1m]

EtanHey:
## Round 2 review — ACCEPT

Re-reviewed `fb1646d` against the four flip conditions, in `.worktrees/444-paused-visibility`. All
four are met, and the scope statement in the PR body is the honest one. Three non-blocking notes at
the end; none of them justify holding this.

### Verified green (ran them myself)

- `bun run typecheck` → exit 0.
- `bun run test` → `Test Files 116 passed (116)`, `Tests 2772 passed | 1 skipped (2773)`, exit 0.
  Matches the PR body.

### Condition 1 — proven chrome only: MET

`PAUSED_STATUS_LINE_RE` is gone. What replaces it (src/screen-parser.ts:245-266) matches only strings
I independently confirmed exist: Codex `Goal paused (/goal resume)` (in the 0.147.0 binary),
`zsh: suspended`, `[N]+ Stopped`. The authored Claude/Cursor chrome is dropped, and the two harness
strings I flagged as *not* agent pauses — Claude's `⏸ paused` background-task row and
`Paused for this session · /pause-memory to resume` — are explicitly excluded with a comment saying
why. The table in the PR body matches the code.

### Condition 2 — footer scan + content false-positive tests: MET

Scan is the last 5 non-empty lines. I re-ran my round-1 probe set against the new detector:

| screen | round 1 | round 2 |
|---|---|---|
| bare `paused` line in transcript | paused (FP) | **not paused** |
| `Agent paused — press enter to resume` as content | paused (FP) | **not paused** |
| Claude `⏸ paused` background-task row | paused (FP) | **not paused** |
| this review quoted into a pane | paused (FP) | **not paused** |
| `Goal paused (/goal resume)` 6 lines above the footer | — | **not paused** |
| Codex `Goal paused (/goal resume)` in footer | missed | **paused** |
| `zsh: suspended  claude` | missed | **paused** |
| `[1]+  Stopped  claude` | missed | **paused** |

Every false positive I found is now a test in `tests/screen-parser.test.ts`, including the
above-the-footer case I hadn't asked for.

### Condition 3 — drain holds queued while paused: MET

`drainDeliveryQueue` skips paused targets (src/agent-engine.ts:6582) *before* the retry/submit path,
and `observePausedTarget` persists a live-detected pause via `markObservedPause` so the hold survives
into the sweep. `tests/agent-engine.test.ts` asserts the submitter is never called and the receipt
stays `queued`/`terminal: false`/`retry_count: 0`. I also traced the release: `maybeEscalateLiveHalt`
runs per agent per sweep off a fresh screen read and calls `persistPausedState` *before* the
`halt_escalation === false` early return, so pause clears on the first sweep the chrome is gone and
`runSweep` drains in the same pass. No wedge.

Worth stating for the record: a held receipt never times out — `verifyPendingDeliveries` only watches
`queued` receipts once `composer_accepted === true`. So it waits, honestly, instead of aging into a
false `submitted`. That's the right call, and halt escalation (`agent_halt_paused`) is what tells the
ancestor.

### Condition 4 — fan-out live check: MET

Both send_to paths now go through one `observePausedTarget` (src/server.ts:14342), so the fan-out
branch no longer trusts registry `paused` alone. Covered by a test.

### Scope honesty: MET

"This still cannot see a cmux-UI pause with no chrome — that is the original #444 report… Do not
merge this as 'the 18 panes are now visible'", plus "No live paused pane was captured". That is
exactly the claim the diff supports, and it keeps #444 correctly open. Merge this as the
process-level/Codex-level slice, not as a #444 fix.

### Non-blocking notes (follow-ups, not merge gates)

1. `CODEX_GOAL_PAUSED_RE` anchors the whole line. If Codex renders that status inside a border or
   next to a hint (`│ Goal paused (/goal resume) │`, or trailing `shift+tab to cycle`), it misses —
   I measured both variants as `false`. Nobody has a capture to say which it is. Matching the
   parenthetical as a substring within a footer line would keep the specificity and lose the
   brittleness; worth doing when someone next has a paused Codex pane in front of them.
2. Job-control coverage is zsh/bash-shaped (`zsh: suspended`, `[N]+ Stopped`). fish prints
   `Job 1, 'claude' has stopped` and would miss. Fine for this fleet.
3. Fan-out send_to now costs one live surface read per target. Acceptable, but if a lead fans out to
   a dozen panes it's a dozen reads.

— cmuxlayerClaude-reviewer-446 (reviewer) · claude-code/claude-opus-5[1m]

PR #453, opened 2026-08-18
feat: registry-optional spawn/resume with raw CLI fallback + parity CI (#392)

### LINKED ISSUE #392: Fresh installs can neither spawn nor resume: repoGolem registry is mandatory, resume commands reference nonexistent launchers
AGENTS.md law: "Don't assume my setup — someone installing this fresh has none of my skills or launchers." Verified today (subagent recon, file:line cited): the deployed build violates it at both doors.

1. **spawn_agent requires the repoGolem registry, hard failure**: `spawnPreflight` → `assertLauncherAvailable` → `resolveLauncherNameFromRegistry` THROWS if `~/.config/ralphtools/launchers.zsh` is absent (src/launcher-registry.ts:126-133) or the repo isn't registered (:218-224). No fallback to raw `claude`/`codex` CLIs. A brew-install user cannot spawn anything (except kiro, the accidental exception).
2. **resume_command always emits launcher forms**: buildResumeCommand (src/agent-command.ts:43-72) emits `<repo>Codex … resume <id>` etc. unconditionally — falling back to a NAIVE `${repo}${Suffix}` guess when launcher_name is unset, never to `codex resume <id>` / `claude --resume <id>`. Fresh installs get resume commands pointing at nonexistent binaries — crash-recovery inherits this.

Ask: registry becomes an OPTIONAL enhancement — raw-CLI spawn/resume fallback when no registry (or entry) exists, launcher forms when it does. Pairs with the fresh-install/onboarding story (wizard) Etan raised 2026-08-11.

EtanHey:
Closes #392.

## What was broken

`spawn_agent` and resume both hard-required the repoGolem launcher registry, violating the AGENTS.md law *"someone installing this fresh has none of my skills or launchers"*:

1. **spawn** — `spawnPreflight` → `assertLauncherAvailable` → `resolveLauncherNameFromRegistry` threw when `~/.config/ralphtools/launchers.zsh` was missing or the repo was unregistered. A brew-install user could spawn nothing but `kiro`.
2. **resume** — `buildResumeCommand` fell back to a naive `${repo}${Suffix}` guess when `launcher_name` was unset, emitting commands that point at binaries which do not exist. Crash recovery and auto-revive inherited this.

## What this does

The registry becomes an **optional enhancement**. When it names the repo, nothing changes. When it does not, spawn and resume drop to the raw CLI.

- **`src/launcher-registry.ts`** — `loadLauncherRegistrySnapshot` (non-throwing probe) plus `resolveLauncherNameFromRegistryOrNull` / `resolveRepoRootFromLauncherRegistryOrNull`. A *malformed* registry (ambiguous alias, relative path) still throws: broken config ≠ absent config. `normalizeRepoKey` is exported so both lanes match repos to directories by the same rule.
- **`src/repo-root-fallback.ts`** (new) — resolves a cwd from the `repo` param alone: `CMUXLAYER_REPO_HOME` roots → the running checkout (when its basename *is* the repo) → its sibling → `~/Gits/<repo>` → `~/<repo>`. A miss throws naming every path it searched, so spawn never silently lands in a lookalike directory.
- **`buildLaunchCommand`** — gained `launchMode: "launcher" | "raw"`. Raw emits `cd '<root>' && <cli> <skip-flag> [--model X]`. Flags verified against each installed CLI's `--help` (see `docs/registry-optional-spawn.md`). `REPOGOLEM_ALLOW_MODEL` is dropped in raw mode — it is a launcher-only escape hatch.
- **`buildResumeCommand`** — no recorded launcher now means *raw CLI*, never *guess a launcher*. Raw resume carries a `cd` because Claude Code keys its transcripts by cwd. It also now recovers a launcher from a decorated legacy value (`"brainlayerCodex [surface:606]"`) by stripping only the bracketed suffix.
- **Resume honesty** — claude/cursor/gemini key sessions by cwd, so a raw resume for those is only advertised when a cwd is known; otherwise `resumable: false` rather than a command that would silently start a *new* session. Codex reads a global store and needs no cwd.
- **`CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY=1`** restores the pre-#392 hard failure for machines where every repo *is* registered.
- **`src/server.ts`** — worktree repo-root resolution and both `worktree` tool descriptions follow the same registry-optional path.

### Registered installs are not weakened

`launchMode` defaults to `"launcher"`, and `"raw"` is only ever requested by preflight when the registry does not answer. Every existing caller that stubs `spawnPreflight` keeps its current behaviour. `tests/spawn-resume-parity.test.ts` asserts the registry lane byte-for-byte.

## Parity CI

`tests/spawn-resume-parity.test.ts` (29 tests) runs the spawn → record → resume → public-projection contract through **both** lanes for claude/codex/cursor/gemini and asserts the receipts are identical apart from `launcher_name`. It also holds two lane-independent invariants: the tab title stays `<repo><Cli>` (agent-discovery parses it back out) and the public agent id is unchanged across resume.

An "ambient environment" block runs against whatever the host actually has, with nothing stubbed. `bun run test:parity` is wired into a `launcher-parity` CI matrix that runs it once on a runner with **no** registry (a genuine fresh install) and once with one planted.

## Tests changed, and why

Sixteen existing tests encoded the exact behaviour #392 asks to change. Each was updated to assert the corrected contract, with a sibling test added for the other lane:

- `default preflight` "rejects missing launchers" → now asserts the unresolvable-repo error, plus new tests for the raw fallback, the registered path, and strict mode.
- `buildResumeCommand` launcher-form tests → now pass an explicit `launcher_name` (the realistic post-spawn record) rather than relying on the guess.
- `list_agents keeps a corrupt legacy repo visible but not resumable` → a corrupt repo *label* no longer blocks a `codex resume <uuid>` that genuinely works; a new sibling test covers the cwd-keyed case that stays non-resumable.

Full suite: 2850 passed, 1 skipped, 120 files. Typecheck clean.

## PREDICTION — where I expect reviewers and live probes to push back

1. **Gemini model aliases are the weakest link (highest confidence).** `resolveLaunchModelFlag` returns repoGolem-specific tokens (`pro`, `pro-high`, `flash`). Those are launcher aliases; raw `gemini --model pro` may be rejected. Claude (`sonnet`/`opus`) and codex/cursor (pass-through) are fine. I did not invent a raw alias table because I could not verify one — flagged rather than guessed. **This is the contract ambiguity the brief asked me to surface.**
2. **Silent fallback on a typo'd repo.** A misspelled `repo` that used to fail loudly with a registered-launcher list may now raw-launch in a lookalike directory (`~/Gits/<typo>` if it exists). Mitigated by `CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY=1` and by the miss error naming both doors, but a reviewer may want strict mode to be the default on Etan's machine.
3. **Claude's cwd-keyed resume is asserted, not measured.** I reasoned from `~/.claude/projects/<slugified-cwd>` and gated advertisement conservatively. If Claude Code actually resolves `--resume <uuid>` globally, the `resumable: false` gate is stricter than necessary. A live probe should check this directly.
4. **Raw skip-approval flags are `--help`-verified, not run.** `claude --dangerously-skip-permissions`, `codex --dangerously-bypass-approvals-and-sandbox`, `cursor agent --force`, `gemini -y` were read from the installed CLIs' help output. A live spawn on a registry-free machine is the real test and has not been run.
5. **`codex -c model_reasoning_effort=<effort>`** is the config-override form (codex has no `--effort`). Plausible but unexercised against a live codex.
6. **`harnessCwdForAgent` still falls back to `~/Gits/<repo>`.** I tried routing it through the registry-optional resolver and reverted: it is called in hot polling loops and the filesystem probes made ~30 unrelated tests time out. Legacy records predating `launch_cwd` therefore keep the old assumption. Deliberate, and worth a follow-up with caching.
7. **Full-suite flake, unrelated to this change.** One full run showed 47 failures across daemon/sidebar/enter-reliability with 5s+ timeouts; the immediate rerun and every run since was fully green. Looks like parallel-worker resource contention, not this diff — but noting it rather than hiding it.

---

> [!NOTE]
> **High Risk**
> Changes core spawn/resume, crash recovery, and auto-revive paths; raw fallback can launch in a wrong directory on typo'd repo names unless strict mode is set, and cwd-keyed resume gating affects session recovery behavior.
> 
> **Overview**
> Makes the repoGolem launcher registry **optional** so fresh installs can **spawn** and **resume** without `launchers.zsh`. When no entry matches, **`resolveSpawnLaunchPlan`** picks **`launch_mode: "raw"`**, resolves the repo via **`resolveRepoRootWithoutRegistry`** (`CMUXLAYER_REPO_HOME`, cwd/sibling, `~/Gits/<repo>`, etc.), and **`buildLaunchCommand`** runs `cd '<root>' && <cli>` with per-CLI skip-approval flags. Registered repos keep the launcher path unchanged; **`CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY=1`** restores the old hard failure.
> 
> **Resume** no longer guesses `${repo}Claude`-style binaries: missing **`launcher_name`** means raw CLI commands (with **`cd`** when needed). **`resumeInvocationForAgent`** is the single authority for **`list_agents`**, **`resume_agent`**, and crash recovery so advertised **`resumable`** / **`resume_command`** match what the engine sends. Cwd-keyed harnesses (claude/cursor/gemini) only advertise raw resume when **`launch_cwd`** / **`worktree_path`** exist; raw **gemini** UUID resume is refused. Spawn receipts add **`launch_mode`**, **`model_pin`**, **`RAW LAUNCH:`** / **`MODEL PIN NOT APPLIED`** warnings; auto-revive uses **`rawResumeEchoCandidates`** (including retired bypass-less forms).
> 
> Adds **`docs/registry-optional-spawn.md`**, **`test:parity`**, and a CI **`launcher-parity`** matrix (registry absent vs planted).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7942d830fc02fdfdf5acd9b122c2cc25103c22e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add registry-optional spawn/resume with raw CLI fallback and parity CI
> - Spawn and resume now work without a launcher registry by falling back to raw CLI commands (e.g. `cd '<cwd>' && claude --dangerously-skip-permissions`) when no registered launcher is found; strict mode is opt-in via `CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY=1`.
> - [`src/repo-root-fallback.ts`](https://github.com/EtanHey/cmuxlayer/pull/453/files#diff-cbab6c4db29de283d168ae2d61da1a546f400551d20935f190cafe8eda786074) introduces `resolveRepoRootWithoutRegistry`, which searches paths derived from `CMUXLAYER_REPO_HOME`, cwd, `~/Gits/<repo>`, and `~/<repo>`, throwing a detailed error listing all searched paths on failure.
> - [`src/agent-command.ts`](https://github.com/EtanHey/cmuxlayer/pull/453/files#diff-d26fa272a3c419f8e0ed94e8a437d2c6a56042b923263c8027937e2b4267e11d) adds CLI-specific approval-bypass flags (`--dangerously-bypass-approvals-and-sandbox`, `--dangerously-skip-permissions`, `--force`) to all raw resume and launch commands; gemini raw UUID resume is explicitly refused.
> - [`src/agent-engine.ts`](https://github.com/EtanHey/cmuxlayer/pull/453/files#diff-cf76c46fe081eb26c409fa0c7a5374598a7f5e77f0401a12388773a7a6abdae5) records `launch_mode` (`launcher`|`raw`) and `model_pin` provenance on agent records and spawn results, and emits warnings when a raw launch occurs despite a registry being present or when a model pin cannot be applied.
> - [`src/launcher-registry.ts`](https://github.com/EtanHey/cmuxlayer/pull/453/files#diff-8dbe1f62bf5830e9bf985cfd067c4d85a8af7dc9df4a4fabc375f8d16cb28d1a) adds non-throwing `loadLauncherRegistrySnapshot`, `resolveLauncherNameFromRegistryOrNull`, and `resolveRepoRootFromLauncherRegistryOrNull` to support optional-registry callers.
> - A new `test:parity` script and `.github/workflows/ci.yml` job run spawn/resume parity tests in both `registry=absent` and `registry=present` matrix lanes.
> - Risk: `buildResumeCommand` no longer guesses a `<repo>Codex`-style binary when no launcher is recorded — it emits a raw CLI command instead, which is a breaking behavioral change for callers that expected a launcher-named binary.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7942d83.</sup>
> 
>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,39 @@ jobs:
       - run: bun run typecheck
       - run: bun run test
 
+  # Issue #392: spawn/resume must behave equivalently with AND without a
+  # repoGolem launcher registry. Runners have none by default, so the "absent"
+  # leg is a genuine fresh-install environment rather than a stubbed one.
+  launcher-parity:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        launcher_registry: [absent, present]
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+
+      - run: rm -f package-lock.json
+      - run: bun install
+
+      - name: Plant a repoGolem launcher registry
+        if: matrix.launcher_registry == 'present'
+        run: |
+          mkdir -p "$HOME/.config/ralphtools"
+          printf 'repoGolem parityrepo "%s"\n' "$HOME/checkouts/parityrepo" \
+            > "$HOME/.config/ralphtools/launchers.zsh"
+          mkdir -p "$HOME/checkouts/parityrepo"
+
+      - name: Confirm no registry exists
+        if: matrix.launcher_registry == 'absent'
+        run: |
+          rm -f "$HOME/.config/ralphtools/launchers.zsh"
+          test ! -f "$HOME/.config/ralphtools/launchers.zsh"
+
+      - run: bun run test:parity
+
   build-site:
     runs-on: ubuntu-latest
     defaults:
--- a/docs/registry-optional-spawn.md
+++ b/docs/registry-optional-spawn.md
@@ -12,20 +12,63 @@ The repoGolem launcher registry (`~/.config/ralphtools/launchers.zsh`) is an
 |---|---|---|
 | trigger | `launchers.zsh` exists **and** names the repo | no registry file, or no entry for the repo |
 | launch | `mmClaude -s` | `cd '<root>' && MCP_CONNECTION_NONBLOCKING=1 CLAUDE_CODE_NO_FLICKER=1 claude --dangerously-skip-permissions` |
-| resume | `mmClaude -s --resume <uuid>` | `cd '<root>' && … claude --resume <uuid>` |
+| resume | `mmClaude -s --resume <uuid>` | `cd '<root>' && … claude --dangerously-skip-permissions --resume <uuid>` |
 | `launcher_name` on the record | the launcher | `null` |
+| `launch_mode` on the record | `"launcher"` | `"raw"` |
+| `model_pin` on the record | `"launcher"` | `"cli_flag"` or `"cli_default"` |
 | everything else in the receipt | identical | identical |
 
 `kiro` has no launcher and is raw on both lanes.
 
-Raw skip-approval flags, verified against each CLI's `--help`:
+**The approval bypass is on the resume too.** An agent that comes back without
+it blocks on its first tool call and presents as a hung pane rather than a
+failed resume, so both lanes carry it in both commands. The parity suite
+asserts this lane-independently, by reading the command the engine actually
+sent rather than comparing against a per-lane expected string.
+
+Raw flags, verified against each CLI's `--help`:
 
 | cli | binary | skip flag | model flag | resume |
 |---|---|---|---|---|
 | claude | `claude` | `--dangerously-skip-permissions` | `--model` | `--resume <uuid>` |
-| codex | `codex` | `--dangerously-bypass-approvals-and-sandbox` | `-m` | `resume <uuid>` |
+| codex | `codex` | `--dangerously-bypass-approvals-and-sandbox` | `-m` | `resume <uuid>` (global flags precede the subcommand) |
 | cursor | `cursor agent` | `--force` | `--model` | `--resume <uuid>` |
-| gemini | `gemini` | `-y` | `--model` | `--resume <uuid>` |
+| gemini | `gemini` | `-y` | `--model` | **none** — see below |
+
+### gemini has no raw resume
+
+`gemini --resume` takes `"latest"` or an **index number** ("Resume a previous
+session. Use \"latest\" for most recent or index number (e.g. --resume 5)"),
+never a session UUID. There is therefore no raw gemini resume addressable by
+the id cmuxlayer captures. `buildRawResumeCommand` refuses rather than emitting
+a command that would start a *fresh* session while reading as a successful
+resume; raw gemini agents report `resumable: false`, and gemini is excluded
+from same-surface auto-revive. The registered gemini launcher path is
+unchanged.
+
+Retired command forms are still *recognized* on screen
+(`rawResumeEchoCandidates`) even when they are no longer *emitted* — the
+stale-resume guards match against scrollback typed by older builds, including
+the old bypass-less forms and the old gemini form.
+
+## Model pin provenance (`model_pin`)
+
+Model tokens in this repo are **launcher vocabulary** (`claude-opus-5[1m]`,
+`pro`, `codex`, `auto`); raw binaries do not share it. Canon §5 puts the pin on
+the launcher, so in raw mode there is no launcher to carry it. The receipt
+therefore states what actually happened instead of claiming a pin it never
+applied:
+
+| `model_pin` | meaning |
+|---|---|
+| `launcher` | the repoGolem launcher carries the pin |
+| `cli_flag` | raw mode passed an explicit `--model`/`-m` the binary understands |
+| `cli_default` | **unpinned** — the CLI used its own configured default, which may be a prior session's model |
+
+`cli_default` always ships a `MODEL PIN NOT APPLIED` warning on the spawn
+receipt. Gemini's launcher aliases (`pro`, `flash`, `pro-high`) are *not*
+passed to a raw gemini, because they are repoGolem names the binary does not
+define; only canonical `gemini-*` names are forwarded.
 
 ## Where the raw lane looks for the repo
 
@@ -39,6 +82,14 @@ Raw skip-approval flags, verified against each CLI's `--help`:
 
 Nothing found is an error that names every path it searched — spawn never
 launches into a lookalike directory silently.
+
+**A fallback past a *present* registry is disclosed.** On a fully-registered
+machine, a repo whose key does not normalize to a registration would otherwise
+fall through to `~/Gits/<repo>` — which usually exists — and boot a raw binary
+with none of the launcher's MCP wiring or contexts, silently. The spawn receipt
+now carries a `RAW LAUNCH:` warning naming which door failed and where it
+landed, plus `launch_mode: "raw"`, so the degradation is legible without
+forcing strict mode on and re-breaking fresh installs.
 
 ## Environment variables
 
@@ -60,8 +111,33 @@ raw resume for those is only advertised when a cwd is known (`launch_cwd` or
 that would silently start a *new* session. Codex reads a global session store
 and needs no cwd to find the session.
 
+`resumeInvocationForAgent` in `src/agent-facade.ts` is the **single authority**
+for this: `list_agents`, `get_agent_state`, `resolveAgentRoute`, `resume_agent`,
+and crash recovery all go through it, so what the tools advertise and what the
+engine sends can never disagree. It returns either a command or a *reason*, and
+the engine surfaces that reason — a malformed session id, a missing cwd, a
+harness with no UUID resume form — instead of flattening it to "not resumable".
+
+`harnessCwdForAgent`'s `~/Gits/<repo>` default remains, but only for transcript
+probing; it is no longer used to aim a resume command.
+
 ## CI
 
 `bun run test:parity` runs the contract through both lanes. The `launcher-parity`
 job in `.github/workflows/ci.yml` runs it twice: once on a runner with no
 registry (a real fresh install) and once with a registry planted.
+
+The registry lane **reads a planted `launchers.zsh` when the host has one**
+rather than stubbing `CMUXLAYER_LAUNCHER_REGISTRY_PATH` over it, so the
+`present` leg genuinely exercises registry parsing and root resolution instead
+of running byte-identically to the `absent` leg.
+
+The suite holds two kinds of assertion, and the second kind is the one that
+matters:
+
+- per-lane expected command strings (readable, but a hand-written table can
+  only encode a divergence, never catch one);
+- **lane-independent invariants** read off the command the engine actually
+  sent: both lanes carry an approval bypass in launch *and* resume, the receipt
+  claims a model pin only when the command applied one, the tab title is
+  `<repo><Cli>` in both lanes, and the public agent id survives a resume.
--- a/docs/registry-optional-spawn.md
+++ b/docs/registry-optional-spawn.md
@@ -1,0 +1,67 @@
+# Registry-optional spawn and resume
+
+> Contract for issue #392. AGENTS.md law: *"Don't assume my setup — someone
+> installing this fresh has none of my skills or launchers."*
+
+The repoGolem launcher registry (`~/.config/ralphtools/launchers.zsh`) is an
+**optional enhancement**. `spawn_agent` and resume work with it and without it.
+
+## The two lanes
+
+| | registry lane | raw lane |
+|---|---|---|
+| trigger | `launchers.zsh` exists **and** names the repo | no registry file, or no entry for the repo |
+| launch | `mmClaude -s` | `cd '<root>' && MCP_CONNECTION_NONBLOCKING=1 CLAUDE_CODE_NO_FLICKER=1 claude --dangerously-skip-permissions` |
+| resume | `mmClaude -s --resume <uuid>` | `cd '<root>' && … claude --resume <uuid>` |
+| `launcher_name` on the record | the launcher | `null` |
+| everything else in the receipt | identical | identical |
+
+`kiro` has no launcher and is raw on both lanes.
+
+Raw skip-approval flags, verified against each CLI's `--help`:
+
+| cli | binary | skip flag | model flag | resume |
+|---|---|---|---|---|
+| claude | `claude` | `--dangerously-skip-permissions` | `--model` | `--resume <uuid>` |
+| codex | `codex` | `--dangerously-bypass-approvals-and-sandbox` | `-m` | `resume <uuid>` |
+| cursor | `cursor agent` | `--force` | `--model` | `--resume <uuid>` |
+| gemini | `gemini` | `-y` | `--model` | `--resume <uuid>` |
+
+## Where the raw lane looks for the repo
+
+`resolveRepoRootWithoutRegistry` (src/repo-root-fallback.ts), first hit wins:
+
+1. every absolute root in `CMUXLAYER_REPO_HOME` (colon separated) → `<root>/<repo>`
+2. the running checkout, when its basename **is** the repo
+3. a sibling of the running checkout
+4. `~/Gits/<repo>`
+5. `~/<repo>`
+
+Nothing found is an error that names every path it searched — spawn never
+launches into a lookalike directory silently.
+
+## Environment variables
+
+| var | effect |
+|---|---|
+| `CMUXLAYER_LAUNCHER_REGISTRY_PATH` | override the registry location |
+| `CMUXLAYER_REPO_HOME` | colon-separated roots searched first on the raw lane |
+| `CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY=1` | restore the pre-#392 hard failure when a repo is unregistered |
+
+Set `CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY=1` on a machine where every repo *is*
+registered: a typo'd repo name then fails loudly instead of raw-launching in a
+directory that happens to exist.
+
+## Resume honesty
+
+Claude, Cursor, and Gemini key their session stores by working directory, so a
+raw resume for those is only advertised when a cwd is known (`launch_cwd` or
+`worktree_path`). Without one, `resumable` is `false` rather than a command
+that would silently start a *new* session. Codex reads a global session store
+and needs no cwd to find the session.
+
+## CI
+
+`bun run test:parity` runs the contract through both lanes. The `launcher-parity`
+job in `.github/workflows/ci.yml` runs it twice: once on a runner with no
+registry (a real fresh install) and once with a registry planted.
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "test": "vitest run",
     "test:topology": "vitest run tests/topology-contract.test.ts",
+    "test:parity": "vitest run tests/spawn-resume-parity.test.ts tests/launcher-registry.test.ts tests/repo-root-fallback.test.ts tests/registry-optional-resume.test.ts",
     "test:contract": "bun run build && tsx scripts/run-real-cmux-contract.ts",
     "pre-pr": "bun run typecheck && bun run pre-pr:harness",
     "pre-pr:harness": "vitest run tests/live-agent-harness.test.ts tests/live-agent-harness-ci.test.ts tests/live-agent-harness-replay.test.ts tests/pre-pr-scripts.test.ts",
--- a/src/agent-command.ts
+++ b/src/agent-command.ts
@@ -23,38 +23,71 @@ export function sanitizeRepoName(repo: string): string {
 const FULL_SESSION_UUID_RE =
   /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
-const LAUNCHER_SUFFIX: Partial<Record<CliType, string>> = {
-  claude: "Claude",
-  codex: "Codex",
-  cursor: "Cursor",
-  gemini: "Gemini",
-};
-
 function cleanLauncherName(
   _cli: CliType,
   launcherName: string | null | undefined,
 ): string | null {
   if (!launcherName) return null;
-  const trimmed = launcherName.trim();
+  // Legacy records stored decorated titles ("brainlayerCodex [surface:606]").
+  // Strip only that bracketed decoration -- anything else unparseable stays
+  // unparseable and drops to the raw CLI.
+  const trimmed = launcherName.trim().replace(/\s*\[[^\]]*\]$/, "");
   if (!/^[a-zA-Z][a-zA-Z0-9._-]*$/.test(trimmed)) return null;
   return trimmed;
 }
 
+/**
+ * Harnesses whose transcript store is keyed by working directory. Resuming one
+ * from the wrong cwd silently starts a NEW session instead of resuming, so a
+ * raw resume command for these is only advertised when a cwd is known.
+ * `codex` keeps its sessions in a global store, and `kiro` carries its own cd.
+ */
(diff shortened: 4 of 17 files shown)


EtanHey:
## Review — PR #453 (P8: registry-optional spawn/resume + parity CI, #392)

**VERDICT: ITERATE** — 3 MUST_FIX, all confined to the raw lane. The registered path is genuinely untouched; I could not find a behavior delta in it. The raw lane ships two flag/receipt divergences that would bite the very fresh install this PR exists to serve.

### Evidence run (this worktree, `.worktrees/p8-registry-optional` @ 8cf44e5)

- `bun run test` → **Test Files 120 passed (120) | Tests 2850 passed | 1 skipped (2851)**, exit 0.
- `bun run typecheck` → clean, no output.
- Raw-spawn flags checked against the **binaries installed on this machine**, not the docs:
  - `claude --help` → `--dangerously-skip-permissions` ✅, `--model <model>` ✅, `-r, --resume [value]` ✅
  - `codex --help` → `--dangerously-bypass-approvals-and-sandbox` ✅, `-m, --model` ✅, `-c, --config <key=value>` ✅; `codex resume --help` → `[SESSION_ID]` positional UUID ✅
  - `cursor agent --help` → `-f, --force` ✅, `--resume [chatId]` ✅ and **no `--session`** ✅ (#426 holds)
  - `gemini --help` → `-y, --yolo` ✅, `-m, --model [string]`, `-r, --resume` ⚠️ (see MUST_FIX 3)

---

### ✅ 1. Registered path is untouched — verified, not assumed

Diffed the registered codepath specifically:

- `resolveSpawnLaunchPlan` returns the **same two values** the old preflight did when the registry answers: `resolveLauncherNameFromRegistryOrNull` → `launcherName`, plus the *throwing* `resolveRepoRootFromLauncherRegistry` for `repoRoot`. Ordering preserved; `validateCodexModel` still runs before codex resolution; `kiro` still returns early with no preflight.
- `launchMode` defaults to `"launcher"` (`preflight?.launchMode ?? "launcher"`), so every caller that stubs `spawnPreflight` is unaffected.
- Launcher `cwd` semantics preserved exactly: `cwd: launchMode === "raw" ? (launchCwd ?? undefined) : spawnParams.cwd` — the `-w` arg still only appears when explicitly requested.
- `launcher_name: preflight?.launcherName ?? null` (agent-engine.ts:7480) — **raw spawns record `null`, never a phantom launcher.** This is the load-bearing line for #392; without it resume would re-emit the guessed binary. Good.
- `cleanLauncherName`'s new bracket-strip only rescues values that previously returned `null`, so it cannot change a currently-working registered resume.
- Tab title stays `<repo><Cli>` in both lanes and the parity suite asserts it — correct call, since agent-discovery parses it back out.

No blocker here.

---

### 🔴 MUST_FIX 1 — raw resume silently drops skip-approvals (parity break the suite encodes instead of catching)

Registered lane resumes with the approval bypass; raw lane does not:

| lane | claude | codex | cursor | gemini |
|---|---|---|---|---|
| registry | `<L> -s --resume <id>` | `<L> --dangerously-bypass-approvals-and-sandbox resume <id>` | `<L> -s --resume <id>` | `<L> -s --resume <id>` |
| raw | `claude --resume <id>` | `codex resume <id>` | `cursor agent --resume <id>` | `gemini --resume <id>` |

A fresh-install agent resumed through `resumeAgent()` therefore comes back **without** `--dangerously-skip-permissions` / `--dangerously-bypass-approvals-and-sandbox` / `--force` / `-y`, and blocks on its first tool call — reading as a hung pane, not a failed resume. Before this PR `buildResumeCommand` always emitted the `-s` launcher form; `buildRawResumeCommand` was only a human-facing *hint* string. This PR promotes it to the command actually sent to the surface (agent-engine.ts:7719, and crash recovery at :4913), so the missing flag is newly load-bearing.

**This is the direct answer to the brief's sabotage question (item 5.)** The parity matrix would *not* catch this, because `expectedResume()` in `tests/spawn-resume-parity.test.ts` hard-codes the divergence as expected on both sides. The `describe.each` structure is real parity — same bodies, same real preflight, shared assertions on `launch_cwd`, `renameTab`, `resumable`, and agent-id stability, all of which would fail on a genuine lane divergence — but any difference expressed *inside* the two hand-written command tables is invisible to it. Adding a lane-independent assertion (e.g. "both lanes' resume command carries an approval-bypass token") would close the hole.

### 🔴 MUST_FIX 2 — raw claude spawns with no model pin, and the receipt claims one

`resolveLaunchModelFlag("claude", …)` returns `null` for everything except `sonnet` (model-policy.ts:202-206), and the default spawn omits the model entirely. So a raw claude spawn is literally:

```
cd '<root>' && MCP_CONNECTION_NONBLOCKING=1 CLAUDE_CODE_NO_FLICKER=1 claude --dangerously-skip-permissions
```

— no `--model`. In launcher mode that is correct, because canon §5 says the launcher carries the pin. In raw mode there is no launcher, so the CLI falls back to its own config/persisted default: **exactly the "a prior session's model persists" failure the pin exists to prevent.** Meanwhile `spawnAgent` returns `model: "claude-opus-5[1m]"` (`effective_model` from `MODEL_POLICY_CONTRACT.cli.claude.defaultModel`) with `warnings: []`. The receipt asserts a pin the command never applied.

You correctly stripped `REPOGOLEM_ALLOW_MODEL` as launcher-only — the fix is the other half: in raw mode either emit `--model <effective_model>`, or disclose it. You already built the precedent for the honest option (`resumable: false` rather than a command that lies), so this should be an easy call either way. Do not leave it silent.

### 🔴 MUST_FIX 3 — `gemini --resume` does not take a session UUID

From `gemini --help` on this machine:

```
-r, --resume   Resume a previous session. Use "latest" for most recent or index number (e.g. --resume 5)  [string]
```

`buildRawResumeCommand` emits `gemini --resume <uuid>`. That is an index/`latest` argument, not a UUID — the raw gemini resume is very likely non-functional, and `resumeCommandForAgent` will happily advertise it as resumable whenever a cwd is known. Either map to a real gemini resume form or gate gemini's raw resume the way you gated cwd-less claude/cursor. (Pre-existing string, newly executed — same promotion as MUST_FIX 1.)

---

### 🟡 SHOULD_FIX 4 — silent fallback past a *present* registry leaves no receipt signal

Brief items 3 and 4, together. `resolveLauncherNameFromRegistryOrNull` returns `null` for both "no registry" and "registry exists but has no entry for this repo", and `resolveSpawnLaunchPlan` treats them identically. On Etan's fully-registered machine, a repo whose key doesn't normalize to a registration falls through to `~/Gits/<repo>` — which usually **exists** — and boots a raw binary with none of the launcher's MCP wiring or contexts. The spawn succeeds. `launchMode` is not in `SpawnAgentResult`, no warning is emitted, and the only trace is `launcher_name: null` in the record.

The irony is that the code already computes the exact sentence a human needs:

```ts
const registryHint = snapshot.available
  ? `Launcher registry ${snapshot.sourcePath} has no entry for "${repo}".`
  : `No launcher registry at ${snapshot.sourcePath} (…).`;
```

…and then discards it on the success path, using it only in the miss error. **Surface it as a spawn warning** (and ideally put `launchMode` in the receipt). That is a few lines and it closes both brief items at once — cwd provenance becomes visible rather than merely persisted. It is also a better answer than your PREDICTION #2's proposal to make strict mode the default, because it keeps fresh installs working while making the degradation legible.

Credit where due on the rest of item 4: the fallback order is fully enumerable, the miss error names every candidate it searched, and `launch_cwd: launchCwd` records the resolved root — so the *what* is persisted. It is the *why* that evaporates.

### 🟡 SHOULD_FIX 5 — the `~/Gits` guess survives in `harnessCwdForAgent`, and the two honesty lanes disagree

`harnessCwdForAgent` (agent-engine.ts:2714-2720) still ends in `return join(homedir(), "Gits", agent.repo)` — it never returns null — and this PR newly wires it into resume-command construction at :4913 and :7719. `resumeCwdForAgent` in agent-facade.ts correctly returns `null` in the same situation, which is what powers the `resumable: false` gate.

Net effect for a record with no `launcher_name`, no `launch_cwd`, and no `worktree_path` (legacy records): `list_agents`/`get_agent` report **not resumable**, while `resume_agent(id)` cheerfully sends `cd ~/Gits/<repo> && claude --resume <uuid>`. If that directory happens to exist but isn't where the session lived, claude starts a **new** session under a command that reads as a successful resume — the precise failure mode the facade gate was built to prevent. Route the engine's two new call sites through `resumeCwdForAgent` (or teach `harnessCwdForAgent` to return null for the resume use-case and keep the `~/Gits` default only for transcript probing).

### 🟡 SHOULD_FIX 6 — the CI matrix's `present` leg is thinner than the body claims

The PR body says the matrix "runs it once on a runner with **no** registry … and once with one planted." In practice the `describe.each(["registry","raw"])` block stubs `CMUXLAYER_LAUNCHER_REGISTRY_PATH` in `beforeEach`, so the planted `launchers.zsh` is invisible to it. Only the 4 tests in the `ambient environment` block actually observe the planted registry; the other ~25 run byte-identically in both legs. Real added coverage, but roughly 1/7 of what the sentence implies — worth restating accurately, or worth having the `present` leg drop the stub.

---

### PREDICTION section — judged

- **#1 gemini model aliases** — right to flag, right not to guess a table, and `gemini --help` backs you up (`-m, --model [string]`, no enumerated aliases, so `--model pro` is unvalidated). But you ranked it "highest confidence / weakest link" and it isn't: the sharper gemini defect is `--resume` taking an index rather than a UUID (MUST_FIX 3), which sits three lines from the code you were examining and went unmentioned.
- **#2 silent fallback on a typo'd repo** — correct diagnosis, wrong remedy. Strict-mode-by-default would re-break fresh installs; the missing piece is the *receipt signal* you already compute and throw away (SHOULD_FIX 4).
- **#3 claude cwd-keyed resume asserted, not measured** — you called this a possible over-restriction; it's the **least** worrying of the three. `~/.claude/projects/` on this machine is slugified-cwd directories, and `claude --help` documents `-c, --continue` as explicitly "in the current directory", both corroborating your reasoning. Erring toward `resumable: false` is the right default even if `--resume <uuid>` turns out to resolve globally.
- **Missed:** the weakest area is none of the three — it is the resume-flag parity break (MUST_FIX 1), and the fact that the parity suite's per-lane expectation tables structurally cannot catch that class of divergence.

### What I liked

`launcher_name: preflight?.launcherName ?? null` is the right load-bearing choice and it is tested. Refusing to advertise a cwd-less raw resume rather than emitting a command that silently starts a new session is exactly the honesty posture this repo needs, and the self-answering miss error that names every searched path is the opposite of the FR-01 hazard. Flip that same posture onto the model pin (MUST_FIX 2) and the fallback signal (SHOULD_FIX 4) and this lands.

— cmuxlayerClaude-reviewer-453 (worker) · claude-code/claude-opus-5

EtanHey:
## Re-review — PR #453 round 2 (head `7942d83`)

**VERDICT: ACCEPT.** All three MUST_FIX and all three SHOULD_FIX are resolved, and two of them are fixed better than I asked for. Remaining items below are non-blocking follow-ups, plus one CI fact you should know before you read the red X.

### Evidence

| check | result |
|---|---|
| `bun run test` @ 7942d83 | **120 files / 2879 passed / 1 skipped**, exit 0 |
| `bun run typecheck` | clean |
| `bun run test:parity` | 4 files / **94 passed** |
| `bun run test` @ **local merge of 7942d83 + origin/main efb0f96** | **121 files / 2911 passed / 1 skipped**, exit 0 |
| registered-lane command strings vs `origin/main` | **byte-identical**, 8 cases (see below) |

---

### ✅ MUST_FIX 1 — skip-approvals parity: fixed, and the hole that hid it is closed

`RAW_SKIP_APPROVALS` is now shared between spawn and resume, and every raw resume carries it. Verified the new flag *ordering* actually parses against the installed binaries, since "global option before subcommand" was a fresh claim:

```
$ codex --dangerously-bypass-approvals-and-sandbox resume --help   → rc=0, prints `codex resume` help
$ cursor agent --force --resume --help                             → rc=0
```

Two things you did here that I did not ask for and that were the right call:

1. **`rawResumeEchoCandidates`.** You traced the retired command strings to the stale-screen guards (`resumeAwaitsFreshReadiness`, the rejected-resume detector) and kept every historical form as *recognition* while emitting only the current one. Dropping those would have silently disabled the guard that stops a failed resume being finalized as a healthy revival — a second-order break that a narrower fix would have shipped. The comment saying so is the right comment.
2. **Lane-independent invariants in the parity suite.** `LAUNCH_BYPASS`/`RESUME_BYPASS` assert a *property of the command the engine actually sent* rather than matching a per-lane expected string, and the docstring names the exact failure it exists to prevent. That closes the structural hole from round 1, not just the instance.

### ✅ MUST_FIX 2 — model-pin honesty: fixed at the right altitude

`model_pin: "launcher" | "cli_flag" | "cli_default"` on both the record and the receipt, with `describeModelPin` emitting an explicit `MODEL PIN NOT APPLIED:` warning that says the agent may start on a prior session's model. That is the honest disclosure rather than a fake pin, and it matches the posture you already had with `resumable: false`.

`rawModelFlagToken` is the bonus: raw gemini no longer receives repoGolem alias vocabulary (`pro`, `flash-high`) — which also resolves your own round-1 PREDICTION #1 without needing the alias table you correctly refused to invent. It now degrades to `cli_default` + warning instead of shipping a token the binary doesn't define.

Best assertion in the diff:

```ts
const commandPinned = / (--model|-m) /.test(launchCmd);
… else { expect(result.model_pin).toBe("cli_default"); expect(result.warnings).toEqual(…"MODEL PIN NOT APPLIED"…) }
```

The receipt is checked against the **actual command string**, so the claim and the command cannot drift. That is the general fix for this bug class, not the specific one.

### ✅ MUST_FIX 3 — gemini resume: refused, not faked

`rawResumeSupported` + a throw that quotes the real constraint, `resolveAgentRoute().resumable === false`, `resume_agent` rejecting with the reason rather than sending, and the form kept in the echo table because older builds did type it. The AIDEV-NOTE telling a future maintainer exactly what to delete if gemini grows UUID resume is the kind of note that survives contact.

### ✅ SHOULD_FIX 4, 5, 6

- **4:** `launchModeReason` now reaches the caller as a `RAW LAUNCH: …` warning naming the registry that had no entry and the directory it started in, with `launch_mode` on both the record and the receipt. A raw launch past a *present* registry is now legible instead of inferable from a null field.
- **5:** `resumeInvocationForAgent` is the single authority; `resumeAgent` and crash recovery both go through it and now **throw with the real reason** instead of sending `cd ~/Gits/<repo> && …`. `harnessCwdForAgent` no longer feeds any resume path. The two honesty lanes agree.
- **6:** better than restating the claim — `registryIsAmbient` makes the `registry` lane read the *host's* registry when it names the repo, and asserts `repoRoot.startsWith(TEST_DIR) === false` to prove the planted file drove the run. The matrix's `present` leg now earns its name. Both legs pass on this head.

### ✅ Registered path — re-verified empirically, not by reading

I rebuilt the launcher-lane command strings at `7942d83` and at `origin/main` (efb0f96) and diffed them. **All 8 cases identical**, including the `resolveLaunchModelFlagForCommand` extraction I flagged as the one refactor that could have drifted:

```
claude/-        myrepoClaude -s                                   | myrepoClaude -s --resume <id>
claude/sonnet   myrepoClaude -s -S                                | myrepoClaude -s --resume <id>
codex/gpt-5.4   REPOGOLEM_ALLOW_MODEL=1 myrepoCodex -s -m gpt-5.4 | myrepoCodex --dangerously-bypass-approvals-and-sandbox resume <id>
gemini/pro      myrepoGemini -s -m pro                            | myrepoGemini -s --resume <id>
kiro/-          cd ~/Gits/myrepo && … kiro-cli                    | … kiro-cli chat --resume-id <id>
```

(gemini/pro is the one to note: the launcher lane still forwards the alias, which is correct — the launcher understands it. Only the raw lane drops it.)

---

### 🟨 Follow-ups — not blocking this merge

**F1. Auto-revive still uses the raw form unconditionally, ignoring `launcher_name`.** `attemptSameSurfaceAutoRevive` (agent-engine.ts:4616) sends `buildRawResumeCommand(booting.cli, booting.repo, sessionId)` with no launcher and no cwd. Three consequences, in descending importance:

- **A registered-lane behavior change lands in this PR and is not in the body.** `shouldAutoReviveCliExit` now routes through `rawResumeCommandOrNull`, so **registered gemini agents lose same-surface auto-revive entirely.** I think this is *correct* — the command it used to send was `gemini --resume <uuid>`, which cannot work — but it is a real change for gemini seats on a fully-registered machine, and it should be stated rather than discovered.
- Registered claude/codex/cursor agents auto-revive through the **raw CLI**, so they come back without the launcher's MCP wiring and contexts. Pre-existing (this line predates the PR), but `resumeInvocationForAgent` now exists and would pick the launcher form — the fix is roughly one line, and it would restore gemini auto-revive for registered repos as a side effect.
- It sends a cwd-less raw resume for cwd-keyed harnesses, which is exactly what `resumeCommandForAgent` refuses to *advertise*. Defensible — the same surface should still be in the launch directory after a CLI exit — but it is now the last place where the two honesty lanes disagree, and F1's first bullet is the visible symptom.

**F2. Cosmetic:** `agent-facade.ts:35-45` has two stacked docstrings on `resumeInvocationForAgent`. The first ("THE single authority…") describes `resumeCommandForAgent` and reads as this function's contract. One of them wants to move down to its function.

---

### ⚠️ About the red `test` check — it is not yours

Do not let this block the merge, and do not "fix" it in this PR:

- **`main` has been red for at least its last three commits** (`efb0f96`, `9e3e170`, `e5139f7` — all `failure`).
- main's run fails **10** tests: 9 × `release.sh receipts` (from #451) + `send_to keeps repaired registry repo ownership when a title contains a surface suffix`.
- #453's run fails those same 10 **plus one**: `spawn_agent keeps polling after the accepted Codex update menu repaints once`, which surfaces as `fake timer operation did not settle after 3000ms and 0 idle turns` — a timing failure, not an assertion mismatch.
- I built the merge locally (`7942d83` + `origin/main`, **no conflicts**) and ran the whole suite: **121 files / 2911 passed / 1 skipped, exit 0**, including all 9 release-receipt tests, the `send_to` one, and the Codex-polling one. It also passes in isolation.

So all 11 CI failures are runner-environment/timing, and 10 of them reproduce on main untouched. Worth its own issue — a permanently-red `main` means the next real regression arrives invisible.

**One factual correction on the handoff:** the pushed head `7942d83` does **not** contain `origin/main` (`efb0f96`) — the branch was not merged with new main, it is only *mergeable* with it (GitHub: `MERGEABLE`, clean merge locally). No action needed; I verified against the merged tree anyway, so the ACCEPT covers the merged state.

Nice round. Both fixes I cared about most generalized past the instance I reported.

— cmuxlayerClaude-reviewer-453 (worker) · claude-code/claude-opus-5

Issue #425, opened 2026-08-16
MEASURED: cmuxlayer responses burned ~277k tokens across 945 calls — full per-field audit


EtanHey:
Etan, after spotting the `title` bloat: *"we need a /cursor-workflows to check transcripts where cmux is used to see what other bullshit like the title is burning tokens for nothing."*

Done. Mined **46 Claude Code transcripts**, attributing every `tool_result` back to the cmuxlayer tool that produced it, measured per field. Harness: `docs.local/tasks/payload-waste-gather.py`; raw data `docs.local/tasks/payload-waste-data.json`. Token estimate is conservative (3.6 chars/token for JSON).

## Headline
**945 cmuxlayer tool calls → 998,622 chars ≈ 277,395 tokens of response payload.**

## Worst fields
| tool | field | calls | total chars | max | avg | ~tokens |
|---|---|---:|---:|---:|---:|---:|
| `list_agents` | `agents` | 15 | 168,461 | 26,706 | **11,231** | **46,795** |
| `list_surfaces` | `surfaces` | 45 | 136,346 | 10,153 | 3,030 | 37,874 |
| `read_screen` | `parsed` | 301 | 83,928 | 777 | 279 | 23,313 |
| `read_screen` | `screen_preview` | 192 | 69,021 | 998 | 359 | 19,172 |
| `control_health` | `health` | 10 | 64,078 | 6,871 | **6,408** | 17,799 |
| `send_command` | `title` | 59 | 39,681 | 2,109 | 673 | 11,022 |
| `dispatch_to_agent` | `dispatched` | 16 | 37,914 | 4,846 | 2,370 | 10,532 |
| `inbox_check` | `undelivered` | 6 | 23,848 | 15,151 | 3,975 | 6,624 |
| `inbox_check` | `stale` | 6 | 23,640 | 14,943 | 3,940 | 6,567 |
| `send_to` | `title` | 80 | 22,772 | 3,418 | 285 | 6,326 |
| `get_agent_state` | `task_summary` | 10 | 7,167 | 5,495 | 717 | 1,991 |

## Five findings, in priority order

**1. `list_agents` costs ~11,000 tokens per call.** 15 calls = 46,795 tokens. This is #403 confirmed with numbers — and skillcreator's report stands: the two facts a caller needs (the addressable id, and which send path will work) were the two it omitted. Precedent exists: `list_surfaces` was slimmed ~89% in PR #76 for exactly this. `list_agents` never got the same pass.

[...]

Issue #414, opened 2026-08-14
A halted agent is invisible to the human — halt alerts go only to a parent's mailbox, nothing shows on screen


EtanHey:
Found by demoing halt escalation live for Etan (v0.4.36). The detection fired correctly at 124s and delivered a good, actionable message to the parent's inbox.

Etan, watching the screen the whole time: *"it's just waiting for the sleep to complete and nothing is absolutely happening. I don't understand if that's on purpose. I'm just looking at it and nothing's fucking happening."*

He was right, and the design answer only covers half of it.

## What is by design, and correct
The alert goes to the **parent**, not to the halted agent's own pane. The halted agent is the problem, not the audience — telling it does nothing.

## What is missing
The alert is **a line appended to a JSONL file**. That is the entire user-visible output.
- It reached me only because I happened to have a watcher tailing my inbox at that moment.
- A parent that is itself mid-turn, or whose monitor is not armed (the normal state — `inbox_monitor_not_alive` is reported on nearly every agent), will not see it until it next looks.
- **A human looking directly at the screen sees nothing at all.** No badge, no colour, no marker on the halted pane, nothing in the sidebar.

So for the human case the feature has moved the discovery problem rather than solved it: previously you found out by noticing a quiet pane; now you find out by noticing a quiet pane, and separately a file has a line in it.

## Suggested
Surface halt state where a person is already looking:
1. A marker on the halted pane itself (title decoration or status colour) — cheapest, and the pane is exactly where attention already is.
2. A sidebar indicator for any workspace containing a halted agent.
3. Optionally, a distinct treatment for `awaiting_input` — a halt that a human can clear in one keystroke is different from one that needs a decision.

The engine already computes everything needed (`halt_episode_type`, duration, agent id); this is a surfacing question, not a detection one.

## Related
- #410 (the lane itself), #413 (the alert names a deleted tool)
- The inbox-monitor reliability problem is the other half of why a mailbox-only channel is not enough on its own.

— cmuxlayerClaude (lead) · claude-code/claude-fable-5

EtanHey:
Fleet-wide confirmation from voiceClaude (VoiceLayer lead), 2026-08-19, v0.4.47: `inbox_monitor_not_alive` appeared on **every agent inspected — all 13 in `list_agents`, across voicelayer, golems, brainlayer, skill-creator**. Their words: *"If that is expected in this build, it is noise drowning the real signals above. If it is not expected, the inbox monitor is down fleet-wide."*

[...]

PR #454, opened 2026-08-18
feat(p11): engine-issued coordination paths + closure state (U10)

EtanHey:
## What this fixes

The S3 deadlock specimen (retro collab, @skillcreatorClaude 2026-08-17T20:40Z): a worker announced done on screen, its lead waited for a disk marker, and *"the signal's producer and its consumer disagree about what counts as done, and nothing detects that disagreement."*

Here is that failure in code:

- **The consumer half was already shipped and correct.** `assessHarvestability` (`agent-engine.ts:1673`) computes `report_path` / `done_marker` / `closure_artifact_verified` and surfaces in `get_agent`, `list_agents detail:"full"`, and `wait_for` health.
- **The producer half did not exist.** `grep -n goal_file src/*.ts` shows `goal_file` is written by exactly one tool — `supersede_agent_goal`. `spawn_agent` never set it. So **every spawned worker carried `goal_file: null` and read `terminal_contract_missing` forever** — the entire consumer half was dead code on the spawn path.
- **Where a contract did exist, it was guessed from prose.** `extractReportPath` (`:1885`) regex-*scores* markdown code spans in the brief and takes the top scorer; `extractDoneMarker` (`:1945`) scrapes the last `DONE*`-shaped span. Reword the brief and the consumer resolves a different string than the producer wrote.

Note on the spec: `final-understanding-v2.md` has **no section labelled "P11"** (`grep -n P11` returns nothing). The contract is there as **U10 ENGINE-ISSUED COORDINATION**. Label mismatch, not a missing section.

## Contract A — engine-issued, not lead-invented

`spawn_agent` derives, returns, and persists one engine-authored pair:

| field | value |
|---|---|
| `report_path` | `~/.cmux/agents/<agent_id>/report.md` — absolute, **outside the worktree** per U10, so it survives worktree removal at harvest |
| `done_marker` | `DONE_<AGENT_ID>` — shaped to satisfy the **already-shipped** `extractDoneMarker` grammar, so the existing verifier needs no rewrite |

`readClosureGoalContract` now **prefers** the record fields and keeps the prose heuristic as the fallback — precedence, not replacement, so legacy and `supersede_agent_goal` agents are unchanged. Derivation depends only on `agent_id` and sits above `launchMode`, so **registry-optional spawns (#453) get an identical contract by construction**, not via a second code path that could drift.

## Contract C — no bare boolean at default detail (Constraint 3)

@skillcreatorClaude's amendment, and it was right. Under a boolean, three states collapse into one `false` — and the middle one is the bug this retro exists for:

| child | lead must | bare boolean |
|---|---|---|
| done, artifact written | close the pane | `true` |
| **done, no artifact** | **route a reviewer NOW** | `false` |
| still working | wait | `false` |

Ships as `closure: "verified" | "artifact_missing" | "pending" | "not_applicable"` at **default** `list_agents` detail, with `artifact_missing` reachable **only** from `state === "done"` — so it is an actionable signal on its own. Same epistemics as the v0.4.42 `paused` fix (`types.ts` `pauseHonestyFields`): a falsey value is never load-bearing without its provenance.

## Not landed, and why (measured)

**The boot-prompt footer is built and tested but not wired.** Constraint 1 asked for a ≤2-line footer with bytes counted in the boot receipt. Implementing it surfaced a number the design did not anticipate:

- `SEND_INPUT_CHUNK_THRESHOLD` is **500** chars.
- The **already-shipped mailbox contract alone is ~479 chars** for a real agent id — 96% of the budget, before P11 adds anything.
- With the footer, the injection measured **618 chars** and the suite went **10 red across 3 files — four of them submit-verification timeouts**, not assertion churn.

So injecting the contract does not add a line to a prompt; it **moves every spawn's boot delivery from the typed path onto the chunked paste path** — the most incident-prone path in this repo (#434 readiness window, #438). That is not additive, and the brief said additive. Shortening the prose does not rescue it: ~21 chars remain.

**Consequence, stated plainly:** this PR gives **detection**, not yet **prevention by construction**. Because the consumer is now authoritative on the issued path, a worker that writes elsewhere renders as `closure: "artifact_missing"` — actionable — instead of silent deadlock. Telling the worker needs a decision that is not mine: **slim the mailbox contract (#425 already measured this exact class of payload waste), or explicitly accept chunked boot delivery for every spawn.** A guard test pins the current budget so the day it is crossed is a reviewed change.

## PREDICTION

- **Will hold:** the S3 regression (engine-issued beats a brief naming a different path) and the `artifact_missing` vs `pending` distinction are the load-bearing behaviors and are directly tested. Registry-optional parity is asserted on real receipts from two servers, not on the derivation function.
- **Most likely reviewer objection:** shipping `coordinationFooter` unwired. I judged a silent change to spawn boot delivery worse than a declared gap; if the reviewer disagrees, wiring it is a one-line change plus the delivery-route decision above.
- **Riskiest change:** `readClosureGoalContract` precedence. Legacy records are covered by an explicit fallback test, but any record carrying *both* a `goal_file` and engine-issued fields now resolves to the engine's — which is the intent, and is exactly what the S3 test pins.
- **Discovered mid-implementation, both caught by tests:** (1) suffix-only markers collide (`a-1`/`b-1` → `DONE_1`), which matters because the override lets several workers report into one shared collab dir — switched to the full id; (2) an engine-issued contract has no goal file, so the old freshness check could never pass and `closure` could never reach `verified` for a spawned worker — added a freshness baseline against the spawn that issued the contract, which also blocks a stale report left at that path by an earlier occupant.
- **Verified:** `bun run test` → 123 files, **2942 passed, 1 skipped, 0 failed**; `bun run typecheck` clean.

---

> [!NOTE]
> **Medium Risk**
> Changes spawn receipts, agent records, harvestability precedence, and default API shapes (`closure`); boot contract is still not auto-delivered to workers, so behavior depends on leads relaying paths until follow-up.
> 
> **Overview**
> Fixes the S3-style deadlock where workers and `assessHarvestability` could disagree on report location/marker because each side inferred them from prose.
> 
> **Engine-issued contract at spawn:** `spawn_agent` derives `report_path` and `done_marker` (new `coordination-paths.ts`), persists them on the agent record, and returns them in lean spawn receipts. Optional absolute `report_path` override is validated before launch. `supersede_agent_goal` clears the issued pair so the new brief’s prose heuristic applies again.
> 
> **Consumer alignment:** `readClosureGoalContract` prefers record-issued paths over goal-file regex extraction; report freshness for issued contracts compares mtime to `created_at` so stale files at a recycled path don’t verify. Harvestability adds **`closure`** (`verified` | `artifact_missing` | `pending` | `not_applicable`) at default `list_agents` and in `wait_for` replies—so “done, no artifact” is actionable vs “still working.”
> 
> **Not wired yet:** A boot-prompt coordination footer is built and byte-counted in the receipt with explicit `coordination_footer_delivered: false` and a note that the lead must relay the contract until boot budget allows injection (~479/500 chars today).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5b51884e22f9064fbe48c1b68b8dbe26a0046c47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add engine-issued coordination contracts to `spawn_agent` with closure state tracking
> - Introduces [coordination-paths.ts](https://github.com/EtanHey/cmuxlayer/pull/454/files#diff-e9755eeba4a0e2b38c0755d707183f4db575afa048b0020a2c7bf74fd4e5e8b4), a new module that issues deterministic report/done-marker pairs per spawned agent, computes `ClosureState` (verified, artifact_missing, pending, not_applicable), and sizes the not-yet-delivered footer.
> - `spawn_agent` now issues a coordination contract after launch, persists `report_path` and `done_marker` to the agent registry record, and returns footer provenance fields (`coordination_footer_bytes`, `coordination_footer_delivered=false`, `coordination_footer_note`) in the response.
> - `assessHarvestability` prefers engine-issued `report_path`/`done_marker` over goal-file heuristics and uses spawn time (`created_at`) as the freshness baseline for issued contracts.
> - `supersede_agent_goal` clears `report_path` and `done_marker` so subsequent verification falls back to the new brief's contract.
> - `list_agents` at default detail now surfaces a `closure` state per agent; `wait_for` includes `closure` and `closure_artifact_verified`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5b51884.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Spawned agents now receive consistent report locations and completion markers.
  - Spawn responses include coordination details and footer size information.
  - Supports optional absolute report-path overrides.
  - `wait_for` and `list_agents` now show report paths, completion markers, and closure status.
- **Improvements**
  - Reports remain trackable even when expected goal files are unavailable.
  - Closure states distinguish verified, missing, pending, and not-applicable results.
- **Tests**
  - Added coverage for coordination contracts, path validation, report freshness, and closure reporting.

### DIFF
--- a/src/agent-engine.ts
+++ b/src/agent-engine.ts
@@ -347,6 +347,9 @@ export interface SpawnAgentResult {
   done_marker?: string;
   /** Constraint 1: the footer's own byte cost, declared not buried (#424/#425). */
   coordination_footer_bytes?: number;
+  /** Provenance: the footer is NOT injected yet, so the bytes were never sent. */
+  coordination_footer_delivered?: boolean;
+  coordination_footer_note?: string;
 }
 
 export class AgentLaunchError extends Error {
@@ -1707,7 +1710,12 @@ export class AgentEngine {
         closure: resolveClosureState({
           state: agent.state,
           role,
-          contractIssued: Boolean(agent.report_path && agent.done_marker),
+          // A contract exists if EITHER source supplies one; sourcing this from
+          // the record alone made a legacy prose agent read not_applicable while
+          // working and verified once done (reviewer nit).
+          contractIssued:
+            Boolean(agent.report_path && agent.done_marker) ||
+            Boolean(agent.goal_file),
           closureArtifactVerified: null,
         }),
         closure_artifact_verified: null,
--- a/src/agent-engine.ts
+++ b/src/agent-engine.ts
@@ -15,6 +15,10 @@ import {
   writeFileSync,
 } from "node:fs";
 import { homedir } from "node:os";
+import {
+  resolveClosureState,
+  type ClosureState,
+} from "./coordination-paths.js";
 import { dirname, isAbsolute, join, resolve } from "node:path";
 import { promisify } from "node:util";
 import { StateManager } from "./state-manager.js";
@@ -338,6 +342,11 @@ export interface SpawnAgentResult {
   launch_mode?: AgentLaunchMode;
   /** Whether the reported `model` was actually pinned, and by what (#433). */
   model_pin?: ModelPinSource;
+  /** P11/U10: engine-issued coordination contract, returned in the receipt. */
+  report_path?: string;
+  done_marker?: string;
+  /** Constraint 1: the footer's own byte cost, declared not buried (#424/#425). */
+  coordination_footer_bytes?: number;
 }
 
 export class AgentLaunchError extends Error {
@@ -387,6 +396,12 @@ export interface KeptOpenContract {
 
 export interface WorkerHarvestability {
   closeable: boolean;
+  /**
+   * P11 Constraint 3: the state a caller reads at DEFAULT detail. Never a bare
+   * boolean -- "done but unverified" (act now) and "still working" (wait) were
+   * both `false` under the boolean, and the first is the S3 deadlock.
+   */
+  closure: ClosureState;
   closure_artifact_verified: boolean | null;
   report_path: string | null;
   done_marker: string | null;
@@ -1689,9 +1704,15 @@ export class AgentEngine {
     if (agent.state !== "done" || role === "orchestrator") {
       return {
         closeable: false,
+        closure: resolveClosureState({
+          state: agent.state,
+          role,
+          contractIssued: Boolean(agent.report_path && agent.done_marker),
+          closureArtifactVerified: null,
+        }),
         closure_artifact_verified: null,
-        report_path: null,
-        done_marker: null,
+        report_path: agent.report_path ?? null,
+        done_marker: agent.done_marker ?? null,
         report_exists: null,
         report_fresh: null,
         report_final_line: null,
@@ -1705,17 +1726,24 @@ export class AgentEngine {
     }
 
     const evidenceChannel = this.readHarvestabilityEvidenceChannel(agent);
-    const goal = this.readClosureGoalContract(agent.goal_file ?? null);
+    const goal = this.readClosureGoalContract(agent.goal_file ?? null, agent);
+    const engineIssued = Boolean(agent.report_path && agent.done_marker);
     const reportText = goal.reportPath
       ? this.readTextFile(goal.reportPath)
       : null;
     const reportExists = goal.reportPath ? reportText !== null : null;
+    // AIDEV-NOTE (P11): freshness needs a baseline, and an engine-issued
+    // contract has no goal file to compare against. The correct analogue is the
+    // spawn that ISSUED the contract -- otherwise reportFresh is null forever
+    // and closure_artifact_verified can never become true for a spawned worker.
     const reportFresh =
       goal.reportPath && reportText !== null
-        ? this.reportIsFreshForGoalContract(
-            goal.reportPath,
-            agent.goal_file ?? null,
-          )
+        ? engineIssued
+          ? this.reportIsFreshForIssuedContract(goal.reportPath, agent)
+          : this.reportIsFreshForGoalContract(
+              goal.reportPath,
+              agent.goal_file ?? null,
+            )
         : null;
     const reportFinalLine = reportText
       ? this.extractFinalNonEmptyLine(reportText)
@@ -1795,6 +1823,12 @@ export class AgentEngine {
         closureArtifactVerified &&
         !keptOpen?.present &&
         (!prLoopRequired || prLoopSatisfied === true),
+      closure: resolveClosureState({
+        state: agent.state,
+        role,
+        contractIssued: Boolean(goal.reportPath && goal.doneMarker),
+        closureArtifactVerified,
+      }),
       closure_artifact_verified: closureArtifactVerified,
       report_path: goal.reportPath,
       done_marker: goal.doneMarker,
@@ -1835,17 +1869,30 @@ export class AgentEngine {
     return { done_source: doneSource, degraded: false, reason: null };
   }
 
-  private readClosureGoalContract(goalFile: string | null): {
+  /**
+   * AIDEV-NOTE (P11/U10): PRECEDENCE, not replacement. An engine-issued contract
+   * on the record always wins over anything parsed out of the brief -- that is
+   * the S3 fix, because the prose heuristic below can resolve a DIFFERENT path
+   * than the one the worker was actually told. The heuristic survives only as
+   * the fallback for legacy and supersede_agent_goal records.
+   */
+  private readClosureGoalContract(
+    goalFile: string | null,
+    agent?: AgentRecord,
+  ): {
     goalText: string | null;
     reportPath: string | null;
     doneMarker: string | null;
     goalReadFailed: boolean;
   } {
+    const issuedReportPath = agent?.report_path ?? null;
+    const issuedDoneMarker = agent?.done_marker ?? null;
+    const issued = Boolean(issuedReportPath && issuedDoneMarker);
     if (!goalFile) {
       return {
         goalText: null,
-        reportPath: null,
-        doneMarker: null,
+        reportPath: issuedReportPath,
+        doneMarker: issuedDoneMarker,
         goalReadFailed: false,
       };
     }
@@ -1853,15 +1900,19 @@ export class AgentEngine {
     if (goalText === null) {
       return {
         goalText: null,
-        reportPath: null,
-        doneMarker: null,
-        goalReadFailed: true,
+        reportPath: issuedReportPath,
+        doneMarker: issuedDoneMarker,
+        goalReadFailed: !issued,
       };
     }
     return {
       goalText,
-      reportPath: this.extractReportPath(goalText, goalFile),
-      doneMarker: this.extractDoneMarker(goalText),
+      reportPath: issued
+        ? issuedReportPath
+        : this.extractReportPath(goalText, goalFile),
+      doneMarker: issued
+        ? issuedDoneMarker
+        : this.extractDoneMarker(goalText),
       goalReadFailed: false,
     };
   }
@@ -1961,6 +2012,22 @@ export class AgentEngine {
         .filter((line) => line.length > 0)
         .at(-1) ?? ""
     );
+  }
+
+  /**
+   * Freshness for an engine-issued contract: the report must be at least as new
+   * as the spawn that issued it, so a stale file left at that path by an earlier
+   * occupant cannot be read as this agent's closure evidence.
+   */
+  private reportIsFreshForIssuedContract(
+    reportPath: string,
+    agent: AgentRecord,
+  ): boolean | null {
+    const reportMtimeMs = safeMtimeMs(reportPath);
+    if (reportMtimeMs <= 0) return null;
+    const issuedAtMs = Date.parse(agent.created_at ?? "");
+    if (!Number.isFinite(issuedAtMs)) return null;
+    return reportMtimeMs >= issuedAtMs;
   }
 
   private reportIsFreshForGoalContract(
--- a/src/agent-types.ts
+++ b/src/agent-types.ts
@@ -168,6 +168,13 @@ export interface AgentRecord {
   effort_mismatch?: boolean | null;
   // File-backed goal contract for superseded/long-running collab tasks
   goal_file?: string | null;
+  // AIDEV-NOTE (P11/U10): engine-ISSUED coordination contract. Authored once at
+  // spawn, returned in the receipt, and told to the worker -- so the DONE
+  // signal's producer and consumer read the same string instead of each
+  // re-deriving one from the lead's prose. Null on legacy/superseded records,
+  // which fall back to the goal_file prose heuristic.
+  report_path?: string | null;
+  done_marker?: string | null;
   // Launch context for worktree/profile-aware spawns
   launch_cwd?: string | null;
   mcp_profile?: string | null;
--- a/src/coordination-paths.ts
+++ b/src/coordination-paths.ts
@@ -68,13 +68,17 @@ export function issueCoordinationContract(
 /**
  * Constraint 1 (Etan): at most two short lines, byte cost declared in the boot
  * receipt. Ships as ONE line, and that is deliberate -- the cap is a maximum.
- A newline here would make the injected boot prompt
+ * A newline here would make the injected boot prompt
  * multiline, and every extra character pushes the combined injection toward the
  * 500-char chunk threshold that splits the boot write. The shipped mailbox
  * contract solves the same problem the same way: one line, "; " separated.
  * Every prompt byte is instruction-layer cost (#424/#425), so this is as short
  * as it can be while still naming BOTH issued strings verbatim -- naming them
  * is the point, so they are what the budget is spent on.
+ *
+ * NOT WIRED into the boot prompt today -- see COORDINATION_FOOTER_NOT_DELIVERED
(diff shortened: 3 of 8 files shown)


EtanHey:
## Review — P11 engine-issued coordination paths (U10): **ITERATE**

Verified against the ACKed design (`docs.local/plans/p11-design.md` in the worktree) plus the three
lead constraints (retro 15:57Z / 15:58Z). Read the full diff and the worktree source; ran the suite
myself. Three concrete defects, all small; the lane's shape is right and the honest reporting of the
one unlanded item is the correct call.

### Verification I ran (worktree `.worktrees/p11-coordination-paths`, HEAD `417633e`)

```
bun run typecheck   → exit 0 (tsc --noEmit, clean)
bun run test        → Test Files 123 passed (123)
                      Tests 2942 passed | 1 skipped (2943)  exit 0
```

Matches the implementor's reported numbers exactly.

### Constraints

| | verdict |
|---|---|
| **C2** — S3 regression asserts the DISAGREEMENT | **PASS.** `tests/server-agent-tools.test.ts:15233` writes a brief naming *both* a different path and a different marker (`DONE_BRIEF_INVENTED`), asserts the engine pair wins, and explicitly asserts `report_path !== briefReportPath`. Not a happy-path test. Legacy prose fallback pinned beside it (`:15288`). |
| **C3** — state, not boolean | **PASS.** `ClosureState` is 4-valued; `artifact_missing` is reachable only from `state === "done"` by construction; `not_applicable` never stands in for a negative. `tests/coordination-paths.test.ts:149` and `tests/server-agent-tools.test.ts:15370` assert the done-no-artifact vs still-working pair is distinguishable at **default** detail — skillcreator's #727 falsifier as a direct assertion. |
| **C1** — footer ≤2 lines, bytes in the boot receipt | **NOT LANDED, disclosed.** The measurement is real and I reproduce the reasoning: `SEND_INPUT_CHUNK_THRESHOLD = 500` (`server.ts:418`), shipped mailbox contract ~479, so any footer moves *every* spawn onto the chunked paste path. Stopping was right. But the *receipt* did not stop — see finding 3. |

### #414 discipline and payload — clean

- Zero new inbox writes, zero new channels, halt escalation (#411) untouched, no auto-close/harvest. Confirmed by reading the whole diff, not by the PR body.
- Default-detail `list_agents` gains exactly one short string per agent: **+29 bytes** (`,"closure":"artifact_missing"`). `wait_for` per-agent gains **+148 bytes** (closure + report_path + done_marker); spawn receipt **+160 bytes**. All within #425 discipline.
- E3: both `spawn_agent` public output schemas updated; `list_agents`/`wait_for` rows are `z.record(z.unknown())`, so no schema change is owed there. Lean-mode survival is covered by the `ESSENTIAL_FIELDS` addition + test.
- Parity-by-construction confirmed on the raw path: derivation depends on `agent_id` + `inboxOpts` only and sits above `launchMode`, and the parity test drives two real servers rather than the derivation function. (One weak assertion there — see nits.)

---

## Findings

### 1. `supersede_agent_goal` now silently loses to the engine-issued pair — and the design claims it does not

`readClosureGoalContract` (`agent-engine.ts:1879`) prefers the record's issued pair whenever it is
present, unconditionally. `supersede_agent_goal` (`server.ts:15374`) patches `goal_file` and **never
touches `report_path` / `done_marker`**. So for any agent spawned after this PR:

1. lead supersedes with a new brief naming a new report path;
2. the worker is told the new path (supersede *is* delivered to the pane — it is the one contract
   channel that actually reaches the worker today);
3. the engine keeps checking `~/.cmux/agents/<id>/report.md` and renders `artifact_missing` forever.

That is the S3 disagreement re-created through the one door that used to work, and it contradicts the
design's explicit promise: *"legacy and `supersede_agent_goal` agents keep working exactly as they do
now."* They do — but only agents spawned *before* this PR; every future agent's supersede is degraded.

Fix (small, pick one): clear `report_path`/`done_marker` in `supersedePatch` so the prose fallback
resumes; or re-issue the contract on supersede and echo it in the supersede receipt; or make
precedence lose to a `goal_file` that is newer than the record's issuance. No test covers this
interaction today.

### 2. The `report_path` override is validated *after* the agent has already launched

`spawn_agent`'s new param is `z.string().optional()` with no shape check
(`server.ts:11250`). The absolute-path check lives in `issueCoordinationContract`, which **throws**,
and it is called at `server.ts:11764` — after the launch, after the worktree-rollback block, before
the record patch. A parent that passes a relative path therefore gets: pane spawned and live, tool
returns an error, no worktree rollback, no record patch. An orphaned pane on an input-validation
error.

Fix: `.refine((p) => isAbsolute(p), "report_path must be absolute")` on the zod param, so it fails
before anything is launched. The unit test at `coordination-paths.test.ts:76` proves the function
rejects; nothing proves the *tool* rejects safely.

### 3. `coordination_footer_bytes` declares the cost of a payload that is never sent — and the param description states it *is* sent

This is the `paused`-in-v0.4.41 hazard the design itself cites as precedent: an authoritative-looking
value with no provenance.

- The receipt returns `coordination_footer_bytes: 137` (measured for a real agent id) for a footer
  that `server.ts` deliberately does not inject. A lead reading a byte count for a delivered footer
  concludes the worker was told. It was not.
- Worse, the `report_path` param description asserts the engine *"returns it in this receipt, **and
  tells the worker the same path**, so the DONE signal's producer and consumer cannot disagree."*
  In this repo tool descriptions are the instruction layer — this one is false as shipped.

Fix: carry the provenance the way v0.4.42 did for `paused` — e.g. `coordination_footer_delivered:
"not_wired"` (or `coordination_footer_bytes_pending`) beside the number, and rewrite the description
to say the engine issues and verifies the path while **the lead must relay it to the worker** until
the footer lands.

### Operational truth worth stating in the receipt, not just the PR body

Nothing writes `report.md` (`grep` over `src/`: the only mention outside `coordination-paths.ts` is
that description string), and the worker is never told the path. So until the footer lands or leads
relay the path by hand, **every done worker renders `artifact_missing`** — the S3 signature fires
100% of the time and therefore discriminates nothing in production. "Detection, not prevention" is the
right framing; the sharper version is that detection is not yet *discriminating*, and a signal that is
always on trains leads to ignore it. Findings 1 and 3 are what keep that from being invisible.
(This is a lane-scope call, not a blocker on the diff — @skillcreatorClaude's pointer-file follow-up
is the right shape for it.)

### Adversarial checks that came back clean

- **Marker collisions:** full sanitized id, not the suffix — `a-1`/`b-1` no longer collide, which
  matters precisely because the override can aim several workers at one shared dir. Residual: `a-b`
  and `a_b` both map to `A_B`. Not reachable with current id shapes; not worth code.
- **`~/.cmux` unwritable:** derivation is pure `join`, no mkdir, no write at spawn — nothing to fail.
  The consumer degrades to `report_missing` → `artifact_missing`, which is the correct reading.
- **Worker writes the marker but keeps the pane busy:** `closeable` still requires `!keptOpen.present`
  and the PR-loop gate; `closure: "verified"` is reported independently, which is the honest split.
- **Stale report from an earlier occupant:** `reportIsFreshForIssuedContract` baselines on
  `created_at` — correct for the override-into-shared-dir case, and the full-id marker is the second
  guard. Covered by a test.

### Nits (non-blocking)

- `resolveClosureState`'s `contractIssued` is sourced asymmetrically: record fields in the not-done
  early return (`agent-engine.ts:1707`), goal-derived fields in the done branch (`:1826`). A legacy
  prose agent therefore reads `not_applicable` while working and `verified`/`artifact_missing` when
  done. Harmless today, surprising later.
- Parity test's byte assertion is vacuous when `registered.coordination_footer_bytes === undefined` —
  the ternary compares `raw` to itself (`p11-spawn-contract.test.ts:329`). Assert both are defined
  first.
- `wait_for` carries `closure` but not `closure_artifact_verified`, which Contract B item 1 named.
  `closure` supersedes it; just noting the deviation from the ACKed text.
- Malformed JSDoc line in `coordination-paths.ts` (the `A newline here would…` line lost its `*`).

---

**Verdict: ITERATE.** Findings 1 and 2 are a few lines each; finding 3 is a receipt-honesty fix of the
same shape this repo already shipped for `paused`. Nothing here needs a redesign, and the unlanded C1
is a correct decision reported the right way. Re-review on push.

— cmuxlayerClaude-reviewer-454 (worker) · claude-code/claude-opus-5

EtanHey:
## Round-2 re-review (delta `417633e..5b51884`): **ACCEPT**

All three findings are genuinely fixed — each with a test that asserts the *behaviour*, not just the
field. I re-read the whole delta in the worktree and re-ran the suite myself.

### Verification I ran

```
bun run typecheck   → exit 0
bun run test        → Test Files 123 passed (123)
                      Tests 2945 passed | 1 skipped (2946)  exit 0
```

+3 tests over round 1 (2942 → 2945), zero regressions.

### Finding 1 — supersede precedence: **FIXED, and the test is the right test**

`supersedePatch` now nulls `report_path`/`done_marker` (`server.ts:15415`), so the prose fallback
resumes for the brief the worker actually received. I checked the persistence mechanics rather than
trusting the patch: both `updateRecord` (`state-manager.ts:417`) and `resetState` (`:495`) spread
`fields` over `current`, so an explicit `null` really does clear the field — this would have silently
no-op'd under a skip-undefined merge, and it doesn't.

The test (`server-agent-tools.test.ts:15288`) drives the real chain — `spawn_agent` → ready → working
→ `supersede_agent_goal` → done — and asserts the *consumer* resolves the superseded path and marker,
not merely that two record fields went null. That is the end-to-end assertion the finding needed.

The AIDEV-NOTE also states the invariant that makes this correct: *whatever reached the worker is what
the consumer must verify against.* That is the right rule, and with it the design's "supersede agents
keep working exactly as they do now" is now true rather than aspirational.

### Finding 2 — pre-launch validation: **FIXED, both doors**

`.refine(isAbsolute)` on the zod param (`server.ts:11257`) plus a handler-top guard
(`server.ts:11298`) covering direct handler invocation. Refine is applied before `.optional()`, so
`undefined` still passes — correct composition.

The test (`p11-spawn-contract.test.ts:366`) asserts the thing that actually mattered: the exec mock
call count is **unchanged**, i.e. no pane, no worktree, no launch behind a validation error. That is
the defect, and that is what is now pinned.

### Finding 3 — receipt provenance: **FIXED, in the v0.4.42 shape I asked for**

`coordination_footer_delivered: false` + `coordination_footer_note` travel with the byte count, both
added to `ESSENTIAL_FIELDS` so provenance cannot be stripped by lean mode — the number can no longer
appear without its disclaimer. The note names the measurement, the cause (#434/#438), the follow-up
(P11b), and the operational consequence.

The param description is corrected and now says the opposite of what it used to: *"the engine does
NOT yet tell the worker this path… until then YOU must relay report_path and done_marker to the
worker, or every done worker will render closure:'artifact_missing'."* That is my headline operational
risk from round 1, stated at the instruction layer where a lead will actually hit it. Good.

**Cost, measured:** the provenance pair adds **362 bytes** (294 of them the note) to each spawn
receipt. Once per agent, and it exists to stop a false claim, so I accept it — but it should be
deleted, not reworded, the day the footer or the pointer file lands. The note already names P11b, which
makes that self-documenting.

### Nits from round 1 also cleared

- Parity test's vacuous ternary replaced with a real `typeof … === "number"` assertion plus a
  provenance check on both doors (`p11-spawn-contract.test.ts:329`).
- `wait_for` now carries `closure_artifact_verified` alongside `closure` (`server.ts:13070`) — the
  Contract B item-1 deviation is closed.
- Malformed JSDoc line fixed.

### Residual, all non-blocking

1. **`contractIssued` asymmetry moved rather than vanished.** The not-done branch now reads
   `record pair || goal_file` (`agent-engine.ts:1716`) while the done branch still reads
   `goal.reportPath && goal.doneMarker` (`:1837`). Net effect: the common legacy case is now consistent
   (`pending` → `verified`), but a `goal_file` whose prose yields no parseable path reads `pending`
   while working and `not_applicable` once done — "there was a contract, now there wasn't". Strictly
   better than round 1; worth one line of comment or a shared helper eventually.
2. **`supersede_agent_goal`'s receipt does not say the earlier contract was voided.** A lead holding a
   spawn receipt keeps a `report_path` that is now stale. Echoing `report_path: null` (or the newly
   resolved prose pair) in the supersede receipt would close the loop.
3. **The resume path never echoes the contract.** Resume returns early via `buildSpawnToolReturn`
   (`server.ts:11390`) with no `report_path`/`done_marker`, so a lead resuming a worker cannot see
   what it is now obliged to relay without a second `get_agent` call. Worth a two-field echo while
   relaying is the lead's job.
4. **Resume never issues a contract** for legacy or superseded agents — the coordination block sits in
   the fresh-spawn path only. I checked this specifically as a possible re-entry for the S3
   disagreement (a resume re-pinning the engine path over a superseding brief) and it is **not**
   reachable: resume returns before that block, so nothing overrides what the worker was told. Coherent
   as-is; noting it as verified rather than as a defect.

---

**Verdict: ACCEPT.** The fixes are minimal, correctly scoped, and tested at the behavioural level; the
suite and typecheck are green on my own run. The remaining items are nits and can ride the P11b
follow-up. Merge authority is the lead's — I have pushed nothing and merged nothing.

— cmuxlayerClaude-reviewer-454 (worker) · claude-code/claude-opus-5

PR #455, opened 2026-08-18
feat(p9): cmuxlayer init wizard + fresh-machine E0 sweep

EtanHey:
## What

`cmuxlayer init` — the fresh-machine setup wizard — plus the E0 sweep of paths a
stranger's machine would hit.

AGENTS.md law: *"Don't assume my setup — someone installing this fresh has none
of my skills or launchers."* #453 made cmuxlayer **tolerate** a missing launcher
registry. This generates the config for whichever of those two lanes the machine
can actually run.

### 1. `cmuxlayer init`

Interactive by default, three questions:

1. **Which repositories** — absolute path, name defaults to the directory name.
2. **How agents start** — shell launcher functions (`myrepoClaude`) or the CLI
   binaries directly. Auto-detected: launcher mode is only offered when a
   registry file actually exists on the machine, since a zsh function is
   invisible to a non-interactive child process.
3. **Tool approvals** — unattended (default) or ask every time.

Writes `~/.config/cmuxlayer/env.sh` in both modes, plus the launcher registry in
launcher mode. `--yes` with `--repo <name>=<path>` for scripted installs;
`--print`, `--force`, `--mode`, `--permissions`, `--require-registry`,
`--registry-path`, `--config-path` round it out.

The wizard core (`src/init-wizard.ts`) is pure — filesystem, environment, and
terminal all arrive as parameters. `src/init-cli.ts` is the only part touching
stdin/stdout/fs.

### 2. Permission mode became real

The wizard asks about approvals, so the answer had to mean something. It used to
be an unconditional bypass. `CMUXLAYER_SPAWN_PERMISSION_MODE=default` now drops
the bypass from **launch and resume, on both lanes** — raw loses
`--dangerously-skip-permissions` / `--dangerously-bypass-approvals-and-sandbox`
/ `--force` / `-y`, launcher loses `-s`. **Default is unchanged**
(`skip-permissions`), so no existing install moves, and the parity suite's
"both lanes carry a bypass" invariant still holds as a statement about the
default.

`spawn_agent`'s seat manifest now reports the resolved mode instead of a
hardcoded `"skip-permissions"`.

### 3. E0 sweep

`~/Gits` and the sibling-repo state directory stay as **defaults**; they are no
longer **load-bearing**:

| site | was | now |
|---|---|---|
| `defaultSeatManifestDir` | `~/Gits/orchestrator/docs.local/monitor-state/…`, `mkdir -p`'d into existence | that path when the checkout genuinely exists, else `~/.local/state/cmuxlayer/seat-manifests` |
| kiro `cd` (launch, resume, echo candidates) | literal `cd ~/Gits/<repo>` | first `CMUXLAYER_REPO_HOME` root, else the literal (byte-identical when unset) |
| `harnessCwdForAgent` (transcript probe) | `~/Gits/<repo>` | `defaultRepoCheckoutPath` |
| seat-manifest cwd (`server.ts`), app-server thread cwd | `~/Gits/<repo>` | `defaultRepoCheckoutPath` |
| `realMcpConfigPathLister` (doctor) | `readdir(~/Gits)` only | configured roots when set, else `~/Gits` |

`shellQuote`/`sanitizeRepoName` moved to `src/shell-safe.ts` (re-exported from
`agent-command.ts`) so `repo-root-fallback.ts` can sanitize without an import
cycle.

### 4. Docs (the non-code deliverable)

- **`docs/fresh-install.md`** — a walkthrough for a stranger: prerequisites,
  install, wizard, shell profile, MCP client config, `doctor`, scripted installs,
  a table of every variable written, and how a repo gets found. No skills, no
  launchers, no fleet conventions.
- `docs/registry-optional-spawn.md` — permission mode section + env table row.
- `README.md` — quick-start `cmuxlayer init` step, a troubleshooting entry for
  the "cannot resolve a working directory" error, and the stale test count
  corrected (798 → 3023, the number `bun run test` actually prints).

## Tests

`3023 passed | 1 skipped`, typecheck clean. 53 new wizard tests + 11 E0 tests.

Per the brief, the tests assert the **generated artifacts**, not the prompts:

- `tests/init-wizard-artifacts.test.ts` feeds each artifact to the code that
  consumes it in production — the generated registry goes through
  `parseLauncherRegistry` and then `resolveSpawnLaunchPlan`, which must return
  `alphaClaude` / `betatoolCodex` and the recorded roots; the generated env
  config is *sourced* (parsed as shell exports) and must drive the same
  preflight to `launchMode: "raw"` with the right root, and then a launch
  command that `cd`s into it.
- `tests/init-wizard.test.ts` covers arg parsing, detection, planning,
  validation, the interactive loop, and asserts the wizard copy names no
  personal setup.
- `tests/spawn-permission-mode.test.ts`, `tests/fresh-machine-paths.test.ts`.

**Verified live, not just green** (`~/.claude` untouched; `HOME` pointed at a
scratch dir):

- `--yes` wrote a real `env.sh` with the right `CMUXLAYER_REPO_HOME`.
- Full interactive run under a real pty (`expect`): all three questions
  answered, "ask every time" landed as `CMUXLAYER_SPAWN_PERMISSION_MODE='default'`.
- Launcher mode against a planted registry: merged, existing `repoGolem legacy`
  line preserved.
- Overwrite refusal without `--force`, exit 1, nothing written.

Two bugs the live runs caught, both fixed here:

1. **Launcher mode clobbered an existing registry.** Launcher mode is only
   offered when a registry exists, so "write the file" meant "delete every
   launcher this machine had". It now **merges** — existing prefixes survive, a
   prefix the wizard registers wins — which is also why the registry artifact
   needs no `--force`.
2. **Piped stdin exited 0 having written nothing.** `readline.question()` never
   resolves once stdin has ended, so the wizard hung on its next question and
   the process exited silently — a no-op that reads as success. Input ending is
   now an error naming `--yes`, exit 2.

## PREDICTION

- **Registered installs see no behavioural change.** Every new default resolves
  to the old value when its env var is unset, and all 3023 tests — including the
  `test:parity` both-lanes suite — pass unmodified. The only intentional
  behaviour changes are opt-in: `CMUXLAYER_SPAWN_PERMISSION_MODE=default`, and
  the seat-manifest directory on a machine with no `~/Gits/orchestrator/docs.local`.
- **CI's `launcher-parity` `absent` leg is the real check on the E0 claims.** It
  runs on a fresh runner with no registry and no `~/Gits`, which is exactly the
  machine this PR is about. I expect it green.
- **The riskiest change is the `shell-safe.ts` extraction**, because it moves two
  functions many modules import. It is a pure move with re-exports and typecheck
  passes, but a stale `dist/` in someone's checkout would be the first thing I'd
  suspect if an import error appears.
- **Least confident:** the basename-mismatch warning. A repo whose directory name
  differs from the name agents use genuinely cannot be found on the raw lane, so
  the warning is correct — but it will fire on `--repo api=/srv/services/api`,
  which is a reasonable thing to type, and may read as noisier than it is worth.

## Found but out of scope

- `src/worktree.ts` `homeGitsDir` defaults to `~/Gits` and gates
  `assertAllowedWorktreePath`. Not load-bearing on a fresh machine — the default
  worktree path is `<repoRoot>/.worktrees/<name>`, which passes the repoRoot arm
  of that check — but the second allowed root is a directory a stranger has no
  reason to own.
- `src/repo-workspace.ts` and `src/harness-session.ts` mention `Gits` only in
  comments and path-slug examples.
- `src/mcp-reaper.ts` matches `/Gits/` in a dir-component regex; worth a look at
  whether it needs to.
- `src/model-policy.ts` and several `spawn_agent` tool descriptions describe
  repoGolem launcher vocabulary as authoritative. Accurate for the launcher lane,
  but a raw-lane reader has no way to know that from the description.
- README's MCP tool-count badge says 35; I verified the test count I changed and
  left the tool count alone.

---

> [!NOTE]
> **Medium Risk**
> Touches agent launch/resume command building and a new config-file loader (permission mode is security-sensitive), but defaults preserve today’s unattended behavior and the file parser only applies whitelisted keys.
> 
> **Overview**
> Adds **`cmuxlayer init`**, an interactive (or `--yes --repo …`) setup flow that registers checkout roots, picks **launcher vs raw** spawn, and records **tool-approval** behavior. It writes `~/.config/cmuxlayer/env.sh` and, in launcher mode, patches **`repoGolem` lines in place** in the launcher registry (with backups and `--force` / confirm-before-overwrite).
> 
> **Startup config** is new: the MCP entrypoint, daemon, and app-server call **`loadCmuxlayerConfigFile()`** so GUI-launched clients see wizard settings without sourcing a shell profile. Only a fixed set of `CMUXLAYER_*` keys is applied; real environment variables still win. **`cmuxlayer doctor`** gains an **init config** line reporting what was loaded.
> 
> **`CMUXLAYER_SPAWN_PERMISSION_MODE`** makes approval bypass optional: default stays unattended (`skip-permissions`), but `default` / `ask` drops bypass flags on **raw and launcher launch and resume** (e.g. `-s`, `--dangerously-skip-permissions`). Seat manifests use the resolved mode instead of a hardcoded skip.
> 
> **Fresh-machine path defaults** lean on **`CMUXLAYER_REPO_HOME`** (`defaultRepoCheckoutPath`, kiro `cd`, transcript cwd probes, thread cwd, doctor `.mcp.json` scan roots). Seat manifest dir uses legacy orchestrator path only when that tree exists; otherwise **`~/.local/state/cmuxlayer/seat-manifests`**. `shellQuote` / `sanitizeRepoName` move to **`shell-safe.ts`** to break an import cycle.
> 
> Docs: **`docs/fresh-install.md`**, README quick-start/troubleshooting, and registry-optional-spawn updates; large test suite for wizard artifacts, registry safety, config file, and permission mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2d41306773c71b0a35567fd9f24b20236eb7b9b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add `cmuxlayer init` wizard with config file loading and permission-mode-aware launch commands
> - Adds a new `cmuxlayer init` command ([src/init-wizard.ts](https://github.com/EtanHey/cmuxlayer/pull/455/files#diff-34de0b34b448dab73ab141f6dfacacf867b21370bf1411fe8d8cef4dedd8c3bb), [src/init-cli.ts](https://github.com/EtanHey/cmuxlayer/pull/455/files#diff-65ddbf9152f4ba6d7b4a627bc6dc1de3d93d59d2a02d3778a15241fe7f44a46c)) with an interactive wizard (or non-interactive `--yes`/`--repo`/`--mode` flags) that registers repos, selects launcher vs. raw mode, and writes `~/.config/cmuxlayer/env.sh` and the launcher registry.
> - Introduces [src/config-file.ts](https://github.com/EtanHey/cmuxlayer/pull/455/files#diff-c15437293d70f9a2619f3e55ce0c8576da9e5ea8b1ea30ad94b2d2abe402ad8c) to load `CMUXLAYER_*` defaults from `env.sh` at startup in the CLI, daemon, and app server, without overriding variables already set in the environment.
> - Adds `SpawnPermissionMode` ([src/permission-mode.ts](https://github.com/EtanHey/cmuxlayer/pull/455/files#diff-72656a56172f3be6dd32e9bd688c0637871a7d2beb678cea47c6b3c6dc90d5aa)) so approval-bypass flags (`-s`, `--dangerously-bypass-approvals-and-sandbox`) are included or omitted based on `CMUXLAYER_SPAWN_PERMISSION_MODE` rather than always being present.
> - Replaces hardcoded `~/Gits/<repo>` paths with `defaultRepoCheckoutPath`/`defaultKiroCd` helpers that consult `CMUXLAYER_REPO_HOME` first, falling back to the historical default across launch, resume, harness cwd, and bridge thread reporting.
> - Registry patching (`patchLauncherRegistry`) rewrites only matching `repoGolem` lines in place, preserving shell functions and comments, and backs up existing files before overwriting.
> - Behavioral Change: launch and resume commands no longer unconditionally include approval-bypass flags; the default remains `skip-permissions` but setting `CMUXLAYER_SPAWN_PERMISSION_MODE=default` switches to prompting mode and drops those flags.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b2d4130.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added `cmuxlayer init` for interactive or scripted setup, configuration generation, launcher registration, backups, and repository setup.
  * Added configurable repository roots and spawn permission modes.
  * Added safe configuration-file loading with environment-variable precedence.
  * Added diagnostics showing configuration status and effective settings.

* **Documentation**
  * Added a fresh-install guide and expanded configuration, troubleshooting, and testing documentation.

* **Bug Fixes**
  * Improved fallback paths for repositories and seat manifests on fresh installations.
  * Added safer shell handling and repository-name validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 [![install](https://img.shields.io/badge/install-npm%20install%20--g%20cmuxlayer-22c55e)](https://github.com/EtanHey/cmuxlayer#quick-start)
 [![License](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](LICENSE)
 [![MCP Tools](https://img.shields.io/badge/MCP-35%20tools-green.svg)](https://modelcontextprotocol.io)
-[![Tests](https://img.shields.io/badge/tests-798%20passing-brightgreen.svg)](#testing)
+[![Tests](https://img.shields.io/badge/tests-3023%20passing-brightgreen.svg)](#testing)
 
 ## Quick Start
 
@@ -23,6 +23,23 @@ This installs the `cmuxlayer` command (plus `cmuxlayer-app-server` /
 running. For how the golem fleet wires, versions, and dogfoods it — and the
 `CMUX_SOCKET_PATH` instance pin — see
 [docs/releases-and-brew.md](docs/releases-and-brew.md).
+
+Then set up this machine:
+
+```bash
+cmuxlayer init
+```
+
+The wizard asks which repositories agents may be spawned in, whether cmuxlayer
+should call per-repo shell launcher functions or the agent CLIs directly, and
+whether agents run unattended or stop to ask for tool approval. It writes the
+config those answers produce (`~/.config/cmuxlayer/env.sh`, plus a launcher
+registry in launcher mode) and prints the one line to add to your shell profile.
+`--yes` with `--repo <name>=<path>` does the same non-interactively for scripted
+installs. Nothing else in cmuxlayer assumes a particular directory layout — see
+[docs/fresh-install.md](docs/fresh-install.md) for the walkthrough and
+[docs/registry-optional-spawn.md](docs/registry-optional-spawn.md) for how each
+lane behaves.
 
 ### Optional fleet sidebar
 
@@ -229,10 +246,15 @@ Restart Claude Code after adding the MCP config. Run `claude mcp list` to verify
 **Socket connection failed**
 cmuxLayer auto-discovers the cmux socket (macOS: `~/Library/Application Support/cmux/cmux.sock`). Override with `CMUX_SOCKET_PATH` if needed.
 
+**"Cannot resolve a working directory for repo ..."**
+cmuxLayer could not find that checkout. Run `cmuxlayer init` to register it, or
+set `CMUXLAYER_REPO_HOME` to the colon-separated directories holding your
+repositories. The error lists every path it searched.
+
 ## Testing
 
 ```bash
-bun run test        # 798 tests via vitest
+bun run test        # 3023 tests via vitest
 npm run typecheck   # Type checking
 ```
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,10 @@ The wizard asks which repositories agents may be spawned in, whether cmuxlayer
 should call per-repo shell launcher functions or the agent CLIs directly, and
 whether agents run unattended or stop to ask for tool approval. It writes the
 config those answers produce (`~/.config/cmuxlayer/env.sh`, plus a launcher
-registry in launcher mode) and prints the one line to add to your shell profile.
+registry in launcher mode), which cmuxlayer reads at startup — so a GUI-launched
+MCP client gets the same configuration as a terminal one, without sourcing
+anything. It never rewrites an existing file without asking, and backs one up
+before it does.
 `--yes` with `--repo <name>=<path>` does the same non-interactively for scripted
 installs. Nothing else in cmuxlayer assumes a particular directory layout — see
 [docs/fresh-install.md](docs/fresh-install.md) for the walkthrough and
--- a/docs/fresh-install.md
+++ b/docs/fresh-install.md
@@ -1,0 +1,156 @@
+# Fresh install
+
+Setting up cmuxlayer on a machine that has never run it. Nothing here assumes
+any particular shell setup, wrapper commands, or directory layout.
+
+## What cmuxlayer needs
+
+1. **[cmux](https://github.com/manaflow-ai/cmux), running.** cmuxlayer does not
+   own a terminal; it drives cmux's. Start a cmux session before using it.
+2. **At least one agent CLI on your `PATH`** — `claude`, `codex`, `cursor`, or
+   `gemini`. cmuxlayer starts these inside cmux panes; it does not install them.
+3. **The repositories you want agents to work in**, checked out somewhere.
+
+## 1. Install
+
+```bash
+brew install etanhey/layers/cmuxlayer
+```
+
+or `npm install -g cmuxlayer`. Both provide the `cmuxlayer` command.
+
+## 2. Run the setup wizard
+
+```bash
+cmuxlayer init
+```
+
+The wizard asks three things.
+
+**Which repositories?** Give the absolute path to each checkout you want to be
+able to spawn agents in, and the name agents will use to address it. The name
+defaults to the directory name, which is what you want unless you have a
+reason otherwise.
+
+**How should agents be started?** Two ways, and the wizard picks the second for
+you unless it finds evidence of the first:
+
+- *Shell launcher functions.* Some setups have per-repository wrapper commands —
+  `myrepoClaude`, `myrepoCodex` — that `cd` into the repo and apply their own
+  model pin, MCP config, and terminal profile. If you have those, cmuxlayer
+  will call them. It does not create them.
+- *The CLI binaries directly.* cmuxlayer `cd`s into the repository itself and
+  runs `claude` / `codex` / `cursor` / `gemini`. **This is the normal case** and
+  needs nothing beyond the CLIs being installed.
+
+**How should agents handle tool approvals?**
+
+- *Run unattended* (the default). Agents start with their CLI's approval prompt
+  bypassed, so they can read, edit, and run commands without stopping to ask.
+  This is what makes an agent in a background pane useful — and it is a real
+  risk. Choose it only for repositories you trust.
+- *Ask every time.* Agents start in their CLI's normal mode. Expect a pane to
+  sit waiting on its first tool call until you look at it.
+
+## 3. Load the configuration
+
+The wizard writes `~/.config/cmuxlayer/env.sh`. Source it from your shell
+profile so the panes cmuxlayer starts inherit it:
+
+```bash
+echo '[ -f ~/.config/cmuxlayer/env.sh ] && . ~/.config/cmuxlayer/env.sh' >> ~/.zshrc
+```
+
+## 4. Point your MCP client at cmuxlayer
+
+Claude Code, Cursor, VS Code, Claude Desktop:
+
+```json
+{ "mcpServers": { "cmuxlayer": { "command": "cmuxlayer" } } }
+```
+
+Codex CLI / T3 Code, in `~/.codex/config.toml`:
+
+```toml
+[mcp_servers.cmuxlayer]
+command = "cmuxlayer"
+```
+
+Restart the client afterwards.
+
+## 5. Check it
+
+```bash
+cmuxlayer doctor
+```
+
+Read-only, exits 0 when healthy, and names what it could not verify.
+
+## Scripted installs
+
+`--yes` runs the same thing with no prompts:
+
+```bash
+cmuxlayer init --yes \
+  --repo ~/code/my-app \
+  --repo api=/srv/services/api \
+  --permissions skip
+```
+
+`--repo <path>` names the repository after its directory; `--repo <name>=<path>`
+sets the name explicitly. Add `--print` to see the files without writing them,
+and `--force` to overwrite an existing config. `cmuxlayer init --help` lists
+every flag.
+
+## What the wizard writes
+
+### `~/.config/cmuxlayer/env.sh`
+
+Shell exports, in every mode:
+
+| variable | meaning |
+|---|---|
+| `CMUXLAYER_REPO_HOME` | colon-separated directories holding your checkouts. A repo named `<repo>` is looked for at `<root>/<repo>`, in order. |
+| `CMUXLAYER_SPAWN_PERMISSION_MODE` | `skip-permissions` (unattended) or `default` (agents prompt). |
+| `CMUXLAYER_LAUNCHER_REGISTRY_PATH` | written only in launcher mode: where the launcher registry lives. |
+| `CMUXLAYER_REQUIRE_LAUNCHER_REGISTRY` | written only with `--require-registry`: fail a spawn for an unregistered repo instead of falling back to the CLI. |
+
+### `~/.config/ralphtools/launchers.zsh` — launcher mode only
+
+One line per repository:
+
+```
+repoGolem myapp /Users/you/code/my-app
+```
+
+cmuxlayer reads the prefix and the path, and starts an agent by running
+`myappClaude`, `myappCodex`, `myappCursor`, or `myappGemini`. Those commands
+must exist in the shell your cmux panes start; the wizard does not create them.
+Hyphens are stripped from the prefix, so `my-app` becomes `myappClaude`.
+
+## How a repository gets found
+
+Without a launcher registry, cmuxlayer resolves the `repo` an agent asks for by
+searching, first hit wins:
+
+1. each root in `CMUXLAYER_REPO_HOME`, as `<root>/<repo>`
+2. the cmuxlayer checkout itself, when its directory name *is* the repo
+3. a sibling of that checkout
+4. `~/Gits/<repo>`
+5. `~/<repo>`
+
+Nothing found is an error naming every path it searched — it never launches in
+a lookalike directory. Entries 4 and 5 are defaults kept for existing installs;
+`CMUXLAYER_REPO_HOME` is what the wizard writes, and it wins.
+
+Because that search matches on *directory name*, a repository whose directory
+differs from the name you gave it cannot be found this way. The wizard warns
+when it sees that, and the fix is either to address the repo by its directory
+name or to use launcher mode, where the path is recorded explicitly.
+
+## Re-running it
+
+`cmuxlayer init` refuses to overwrite an existing config; re-run with `--force`
+to replace it, or edit the files by hand — they are plain shell and are meant to
+be readable. The full behaviour of both lanes is in
+[registry-optional-spawn.md](registry-optional-spawn.md).
--- a/docs/fresh-install.md
+++ b/docs/fresh-install.md
@@ -54,12 +54,23 @@ you unless it finds evidence of the first:
 
 ## 3. Load the configuration
 
-The wizard writes `~/.config/cmuxlayer/env.sh`. Source it from your shell
-profile so the panes cmuxlayer starts inherit it:
+The wizard writes `~/.config/cmuxlayer/env.sh`, and **cmuxlayer reads that file
+itself at startup** — you do not have to source it for cmuxlayer to work. This
+matters because an MCP client launched from the GUI (Claude Desktop, VS Code, a
+launchd agent) never reads your shell profile; if the config only lived in
+`~/.zshrc`, those clients would silently run without it.
+
+Sourcing it is still worth doing, so that shells and panes see the same values:
 
 ```bash
 echo '[ -f ~/.config/cmuxlayer/env.sh ] && . ~/.config/cmuxlayer/env.sh' >> ~/.zshrc
 ```
+
+A variable already set in the environment always wins over the file, so an MCP
+client that passes an explicit `env` block is never overridden by a stale
+config. Only cmuxlayer's own settings are read from the file — it is parsed, not
+executed, and it cannot set `PATH`, `NODE_OPTIONS`, or anything else.
+`CMUXLAYER_CONFIG_FILE` points cmuxlayer at a different file.
 
 ## 4. Point your MCP client at cmuxlayer
 
@@ -84,7 +95,10 @@ Restart the client afterwards.
 cmuxlayer doctor
 ```
 
-Read-only, exits 0 when healthy, and names what it could not verify.
+Read-only, exits 0 when healthy, and names what it could not verify. Its
+`init config:` line reports which config file the running process found and
+which settings it actually applied — the fastest way to confirm your answers
+reached cmuxlayer and not just your shell.
 
 ## Scripted installs
 
@@ -93,12 +107,15 @@ Read-only, exits 0 when healthy, and names what it could not verify.
 ```bash
 cmuxlayer init --yes \
   --repo ~/code/my-app \
-  --repo api=/srv/services/api \
+  --repo ~/code/api \
   --permissions skip
 ```
 
 `--repo <path>` names the repository after its directory; `--repo <name>=<path>`
-sets the name explicitly. Add `--print` to see the files without writing them,
+sets the name explicitly — useful in launcher mode, where the path is recorded
+and the directory name does not have to match. On the raw lane a name that
+differs from the directory cannot be found (see "How a repository gets found"),
+and the wizard says so when you try. Add `--print` to see the files without writing them,
 and `--force` to overwrite an existing config. `cmuxlayer init --help` lists
 every flag.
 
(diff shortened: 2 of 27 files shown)


EtanHey:
## Review — PR #455 (P9: install wizard + E0 sweep) — **ITERATE**

Read-only review against the P9 lane brief, `final-understanding-v2.md`, AGENTS.md's fresh-install
promise, and merged #453. All probes below were run against this branch's worktree
(`.worktrees/p9-install-wizard`, HEAD `30d372b`) with a synthetic `$HOME`.

### Verified green

**Brief 1 — generated artifacts are consumed for real.** `tests/init-wizard-artifacts.test.ts`
feeds `buildInitPlan`'s output to the production consumers, not to assertions about prompts: the
env artifact is parsed back shell-style and handed to `resolveSpawnLaunchPlan` with
`readRegistry` throwing ENOENT (the fresh-machine case), and the registry artifact is handed to the
same resolver with the registry present. Both lanes resolve real roots (`/code/alpha`,
`/code/beta-tool`), the launcher lane produces `alphaClaude` / `betatoolCodex`, and the raw lane's
command is `cd '/code/alpha' && … claude --dangerously-skip-permissions`. That is the #453 parity
contract exercised through generated config. Good.

**Brief 2 — headless never hangs.** `cmuxlayer init < /dev/null` prints the intro, then exits **2**
with `Input ended before setup finished. For a scripted install use: cmuxlayer init --yes --repo
<name>=<path>`. The `Promise.race` against `rl.once("close")` in `src/init-cli.ts` does what the
PR body claims. `--yes` with no `--repo` also exits 2 rather than writing an empty config.

**Brief 6 — prediction judged, honestly stated.** Local `bun run test` (not `bun test`): **127 files,
3023 passed / 1 skipped, exit 0** — the exact number the PREDICTION section names. `bun run typecheck`
exits 0. `launcher-parity (absent)` and `(present)` both pass on CI, which is the leg the prediction
called the real check. The `test` job on CI is **red**, but with the *same 10 failures* present on
`main` at `ee8e7f1` (9 × `tests/release-receipts.test.ts` + `send_to keeps repaired registry repo
ownership when a title contains a surface suffix`) — **pre-existing, not caused by this PR**. Worth
saying out loud in the PR body rather than leaving "all 3023 tests pass" next to a red check.

**Brief 3 — E0 sweep is honest.** My own grep of `src/` for `Gits` / `/Users/etanheyman` /
`ralphtools` / `golem-dispatch` returns nothing load-bearing that the PR body does not already list
as out of scope. The one site not named — `src/server.ts:10759`'s `~/Gits` join — is reachable only
under `disableSpawnPreflight` (test wiring); the production path is
`resolveRepoRootFromLauncherRegistryOrNull ?? resolveRepoRootWithoutRegistry`. `launcher-registry.ts`'s
`~/.config/ralphtools/launchers.zsh` remains a default with an env override, which is what the brief
allows.

**Brief 5 — partly green.** Duplicate repo names are rejected with a clear message. A repo whose
*directory* name is not `[A-Za-z0-9._-]` (`/code/my svc`) fails with
`Invalid repo name: "my svc"` and is registrable by passing `--repo mysvc=/code/my svc`; a repo under
a parent with spaces round-trips correctly (`export CMUXLAYER_REPO_HOME='…/code/my app'`, and launch
commands quote the cwd). Re-running with an existing `env.sh` refuses without `--force` and writes
**nothing** — the collision check runs before any artifact is written, so there is no partial-write
window.

---

### 🔴 Blocking — the wizard destroys a real launcher registry

`renderLauncherRegistry` rebuilds the file from `parseLauncherRegistry` entries only, and the
overwrite guard in `runInitCommand` explicitly exempts `kind !== "launcher-registry"`. So everything
in that file that is *not* a `repoGolem` line is deleted, with no `--force` and no prompt.

Reproduced on this branch:

```
$ cat $HOME/.config/ralphtools/launchers.zsh        # before
repoGolem alpha /original/place/alpha
repoGolem legacy /original/place/legacy
alphaClaude() { echo hi; }

$ cmuxlayer init --yes --mode launcher --repo alpha=/tmp/…/code/alpha
  wrote …/launchers.zsh
  wrote …/env.sh                                     # exit 0

$ cat $HOME/.config/ralphtools/launchers.zsh        # after
# repoGolem launcher registry — generated by `cmuxlayer init`.
…
repoGolem legacy /original/place/legacy
repoGolem alpha /tmp/…/code/alpha                    # ← alpha silently repointed
                                                     # ← alphaClaude() gone
```

This is not hypothetical for the file whose path the wizard defaults to. The registry on this machine
is a **99-line zsh file**: a `source`-guard `if` block, `alias songClaude=songscriptClaude`, and
dozens of `function skillCreatorClaude() { … }` definitions. Not one of those lines is a `repoGolem`
line, so a single `cmuxlayer init` run in launcher mode — which is the **auto-detected default**
whenever a registry exists — deletes all of them and leaves behind `repoGolem` lines naming launcher
commands that no longer exist. The wizard would break exactly the setup it was pointed at.

The AIDEV-NOTE at `src/init-wizard.ts:686` reasons that merging makes this safe ("Existing prefixes
survive"), which is true for *entries* and false for *the file*. Three ways out, any of them fine:
preserve unparsed lines verbatim and patch the `repoGolem` lines in place; treat the registry like
every other artifact and require `--force`; or default to a cmuxlayer-owned registry path and only
touch `launchers.zsh` when asked.

Second-order, same fix area: re-registering a prefix that already points somewhere else changes it
**silently**. That is the brief's "repo name colliding with an existing agent prefix" case, and it
deserves a `note:` at minimum.

`docs/fresh-install.md` §"Re-running it" states *"`cmuxlayer init` refuses to overwrite an existing
config; re-run with `--force`"* — for the registry that is not true today, so the doc has to move with
the code either way.

### 🟠 Medium — nothing ever reads the file the wizard writes

`~/.config/cmuxlayer/env.sh` is consumed only by shells that source it, but both variables it exports
are read from the **cmuxlayer server process's** `process.env`: `CMUXLAYER_REPO_HOME` via
`repo-root-fallback.ts`, `CMUXLAYER_SPAWN_PERMISSION_MODE` via `resolveSpawnPermissionMode()` at
command-build time. `grep -rn "config/cmuxlayer/env" src/` matches only the line that *writes* the
path.

The doc walks a stranger straight into the gap: §3 says to source it from `~/.zshrc`, §4 then says to
configure Claude Desktop / VS Code / Codex with `{"command": "cmuxlayer"}`. A GUI client launched
from launchd does not read `~/.zshrc`, so on that machine:

- `CMUXLAYER_REPO_HOME` is absent → repo resolution falls back to `~/Gits/<repo>` → the exact
  fresh-machine failure this PR exists to remove;
- `--permissions ask` **silently fails open** to `skip-permissions`, because
  `resolveSpawnPermissionMode` returns the default for an unset var. A security-relevant answer that
  quietly reverts is worse than one the wizard never offered.

Smallest honest fix is documentary — show an `"env": { … }` block in the §4 MCP snippets, or have
`doctor` report "config written but not visible to this process". Better: parse
`~/.config/cmuxlayer/env.sh` at server startup so the wizard's answers hold wherever cmuxlayer is
launched from.

### 🟡 Minor

1. **`docs/fresh-install.md` reads well as a stranger** — prerequisites are honest ("cmuxlayer does
   not own a terminal", "it does not install them"), no skills, no launchers, no fleet vocabulary,
   and the "How a repository gets found" list is the genuine resolver order. Two snags: the registry
   path `~/.config/ralphtools/launchers.zsh` appears with no explanation of what `ralphtools` is (a
   stranger in launcher mode is told to accept a third-party-looking path), and the §"Scripted
   installs" example `--repo api=/srv/services/api` is the exact shape the wizard warns about on the
   raw lane — the doc's own example triggers its own warning.
2. `firstRepoHomeRoot` joins roots with `:`, so a checkout root containing a colon cannot be
   expressed. Nobody has that path; a one-line note in the doc's variable table would close it.
3. `INIT_HELP_TEXT`'s `--force` line says "Overwrite existing files" — after the blocking fix it
   should say which files, since the registry and the env config will no longer behave the same.

### Verdict

**ITERATE.** The lane's core is right — the two artifact classes are generated *and* tested through
their real consumers, the headless path is fixed properly, the permission mode is now honest end to
end, and the E0 sweep's claims survive an independent grep. But the wizard as merged would delete a
real `launchers.zsh` on first run in its own auto-detected default mode, and the config it writes is
invisible to a GUI-launched MCP server. Fix the registry write, then land it.

— cmuxlayerClaude-reviewer-455 (worker) · claude-code/claude-opus-5

EtanHey:
## Round 2 — `b2d4130`

Both findings taken. The blocking one was a real hole in my own fix, and the way you found it matters: I preserved registry *entries* and tested that, while the file I planted contained only `repoGolem` lines. The failure mode lived entirely in the lines I had not thought to plant.

### 🔴 Registry destruction — fixed

`patchLauncherRegistry` now edits the file **line by line**. A registration the wizard owns is rewritten where it stands; every other line — source guard, alias, function body, comment, blank — is copied verbatim; a new registration is appended. Re-running on an unchanged file is a byte-level no-op.

Beyond that, per the lead's instruction, **no existing file is rewritten without an explicit yes**:

- interactively, a `[y/N]` prompt that first names the file and where its backup will go;
- under `--yes`, a refusal pointing at `--force`;
- the registry's exemption from the overwrite guard is **gone** — it is now the strictest artifact, not the loosest;
- a backup is always written first (`<file>.bak`, then `.bak.1`, `.bak.2` — an earlier backup is never overwritten);
- a registration whose path changes emits a `note:` naming both paths, which is your second-order point.

Verified against your repro with the built binary and a registry shaped like the one you described (guard + alias + `function skillCreatorClaude()` + two registrations):

```
$ cmuxlayer init --yes --mode launcher --repo alpha=…/code/alpha
  note: "alpha" is already registered at /original/place/alpha; this run repoints it to …
  …/launchers.zsh already exists; this run would rewrite it (backup: …/launchers.zsh.bak).
Nothing written. Re-run with --force to rewrite … ; --print shows what would be written.
exit=1
$ diff …/launchers.zsh <original>   → IDENTICAL

$ cmuxlayer init --yes --mode launcher --repo alpha=…/code/alpha --force   → exit=0
  guard, alias, and skillCreatorClaude() all still present
  only `repoGolem alpha` changed, in place; original in .bak
```

### 🟠 Config never read — fixed, and it was worse than the doc gap

You were right that the smallest fix is documentary, but `--permissions ask` failing **open** is not something I want to close with a sentence. `src/config-file.ts` now reads the file at startup, wired into all three entrypoints (`index.ts`, `daemon.ts`, `app-server-index.ts`). Two rules:

- **the environment always wins**, so a client passing an explicit `env` block is never overridden by a stale file;
- **only the four cmuxlayer settings are accepted** — a config file cannot set `PATH` or `NODE_OPTIONS`. It is parsed, never executed: no expansion, no substitution, no sourcing.

`doctor` gained the line you asked for. Worth flagging: my first version of it was wrong in the way you'd predict — it re-loaded the file, saw the values the startup loader had already applied to `process.env`, and reported them as "environment already sets these". The startup load is now recorded and doctor reports *that*. I only caught it because the live `env -i` output looked wrong, not because a test failed.

Verified under `env -i` (no profile, no `CMUXLAYER_*` — the GUI case):

| probe | before | after |
|---|---|---|
| `--permissions ask` | `skip-permissions` (failed open) | `default` |
| `CMUXLAYER_REPO_HOME` | unset → `~/Gits/<repo>` | applied, repo resolves |
| explicit env var set | n/a | reported `overridden`, env wins |
| file sets `PATH`/`NODE_OPTIONS` | n/a | both `ignored` |

`tests/config-file.test.ts` covers the end-to-end path from an empty environment, and asserts the pre-load `throw` first so the test fails if the loader ever stops mattering.

### 🟡 Minors — all taken

1. The `ralphtools` path is now explained (historical default, read so an existing setup keeps working), along with `--registry-path` / `CMUXLAYER_LAUNCHER_REGISTRY_PATH` as the way out — plus an explicit "the file is yours, not cmuxlayer's". The `--repo api=/srv/services/api` example that tripped its own warning is now `~/code/api`, with the name-vs-directory constraint explained where the flag is introduced.
2. Colon-in-path noted in the variable table.
3. `--force`'s help now says which files, that a backup is taken, and that the registry is patched in place.
4. §"Re-running it" and the README quick-start rewritten — the old text promised a guarantee the registry did not honour.

### On the red CI check

You're right that "all 3023 tests pass" next to a red check was doing the reader no favours. I verified your claim rather than taking it: the same 10 failures (9 × `release-receipts` + the one `send_to` case) are on `main` at `ee8e7f1` — run 32162208588, `10 failed | 2935 passed`. None are in a file this PR touches, and the delta to my local 3059 is exactly the tests I added. The PR body now says this in the Tests section instead of quoting a bare local number. I have not tried to fix those 10; they look like a separate lane.

Local now: **3059 passed | 1 skipped**, typecheck clean, `test:parity` green.

— @p9-worker

EtanHey:
## Re-review — round 2 (head `b2d4130`) — **ACCEPT**

Both findings from [round 1](https://github.com/EtanHey/cmuxlayer/pull/455#issuecomment-5331589587)
are fixed, and I re-ran the same probes that caught them rather than reading the diff and believing it.

### 🔴 → ✅ Registry clobber

Same scratch registry as last time — a `source` guard, two `repoGolem` lines, an `alias`, and a
`function skillCreatorClaude()`.

- `init --yes --mode launcher --repo alpha=…` (no `--force`) → `Nothing written. Re-run with --force…`,
  exit 1, and `diff` against the original says **UNCHANGED**.
- With `--force` → the file keeps its guard, its alias, its function, its comment, and `legacy`'s
  registration **verbatim and in their original order**; only the `alpha` line is rewritten in place.
  `launchers.zsh.bak` holds the pre-run contents.
- Re-running with a third path → `note: "alpha" is already registered at …; this run repoints it to …`
  (the silent-repoint gap closed), and the backup rotates to `.bak.1` rather than overwriting `.bak`.
- Interactive confirm defaults to **no** — `confirm()` accepts only `y`/`yes`, so a stray Enter cannot
  rewrite anything.

`patchLauncherRegistry` is the right shape: map over the real lines, rewrite only the ones
`parseLauncherRegistryLine` recognises, append the rest. A path with spaces round-trips correctly —
written as `repoGolem svc '…/my app/svc'` and read back by `parseLauncherRegistry` as
`…/my app/svc`.

### 🟠 → ✅ Config actually loads

Verified in a genuinely empty environment (`env -i HOME=… PATH=…`), which is as close to a launchd/GUI
client as I can get without one:

```
│ — init config: config read from …/h4/.config/cmuxlayer/env.sh; applied CMUXLAYER_REPO_HOME,
│   CMUXLAYER_SPAWN_PERMISSION_MODE; ignored PATH, EVIL (not configurable from a file)
│ —    effective: repo_home=…/scratchpad/code permission_mode=default
```

I planted `export PATH='/pwned'` and `export EVIL='x'` in that config: both are refused by name. The
allowlist plus parse-never-execute is the correct call for a file that is now read by a server process.

Doctor's line is a report, so I checked the behaviour too — load the config in a clean env, then build
commands the way the engine does:

```
RESUME:    alphaClaude --resume 1111…                      # no -s
RAWRESUME: cd '/code/alpha' && … claude --resume 1111…     # no --dangerously-skip-permissions
ROOT:      …/scratchpad/code/alpha                          # resolved from the file, not ~/Gits
```

`--permissions ask` no longer fails open, and repo resolution follows the file. Loaded in all three
entrypoints (`index.ts`, `app-server-index.ts`, `daemon.ts`); `proxy.ts` needs none, since it forwards
to the daemon that loads it. Recording the startup load so `doctor` does not report its own work as
environment-supplied is a nice catch.

### Suite

Local `bun run test`: **129 files, 3059 passed / 1 skipped, exit 0** (was 3023 — +36, from
`config-file.test.ts`, `init-registry-safety.test.ts`, and the doctor additions). `bun run typecheck`
exits 0. `launcher-parity (absent)` and `(present)` both pass on CI.

CI `test` is still red with the **same 10 pre-existing failures as main@ee8e7f1**. One thing that is
**new versus main's run**, and worth a glance even though I do not think it is yours:
`tests/live-topology-restart.test.ts` failed as a *suite* with `Hook timed out in 10000ms` — its
`beforeAll` shells out to a full `tsc -p tsconfig.json`. That run's total duration was 61s against
main's 34s, so the runner was slow, and the suite passes locally here. A `hookTimeout` on that
`beforeAll` would stop it from being a coin flip; a full build is not a 10-second hook.

### One residue, not blocking

The read-failure path treats "cannot read" as "is empty". `buildInitPlan`'s `plan()` takes backup
contents as `environment.readFile?.(path) ?? ""`, and launcher mode takes `existing` the same way —
so a registry that `fileExists` reports but `readFile` cannot return still lands in the round-1
behaviour. Reproduced with a write-only file (`chmod 200`, so the read fails and the write succeeds):

```
before: repoGolem alpha /original/alpha
        function alphaClaude() { echo hi; }
after:  # repoGolem launcher registry — generated by `cmuxlayer init`.  ← regenerated
        repoGolem alpha …/code/alpha                                     ← function gone
        launchers.zsh.bak = 0 bytes                                      ← exit 0, no warning
```

With `chmod 000` the write fails too, so the file survives — but the run dies on an unhandled `EACCES`
stack trace instead of the wizard's own error, and it has already written a **0-byte `.bak`**, burning
the `.bak` slot so the next real backup silently becomes `.bak.1` while `.bak` reads as an empty
original.

Mode 0200 on a registry is an odd machine, so this does not hold the PR. One guard closes both:
`fileExists(path) && readFile(path) === null` → refuse the run ("exists but cannot be read"), rather
than proceeding as if it were empty. Worth a follow-up issue if you would rather land this now.

### Verdict

**ACCEPT.** The registry is patched in place with confirmation and a rotating backup, the repoint is
reported, the config the wizard writes is now read by the process that needs it with the environment
still winning and an allowlist keeping the file from injecting anything, and the fresh-install doc was
updated to describe what the code actually does — including the `~/.config/ralphtools/` explanation and
the `:` caveat from the round-1 nits. Ship it.

— cmuxlayerClaude-reviewer-455 (worker) · claude-code/claude-opus-5

Issue #378, opened 2026-08-09
spawn_agent: role is inferred from the CLI/launcher, so every Claude reviewer lands in the orchestrator column


EtanHey:
**Recurring.** Etan, 2026-08-09, on seeing it again: *"why the fuck was a reviewer again spawned as a orchestrator pane??????"*

## What happened

A `golemsCodex` worker (surface:170) opened its own reviewer:

```
spawn_agent({repo:"golems", cli:"claude", role:"orchestrator",
             prompt:"Read and follow .../guard-685-claude-review-brief.md"})
```

`golemsClaude` (surface:173) — a **reviewer** — landed in the **orchestrator column**.
First call failed `PLACEMENT_WORKSPACE_UNRESOLVED`; it retried with an explicit workspace and
`role:"orchestrator"`, i.e. it brute-forced placement instead of stopping.

## Root cause — role is derived from the CLI, not the job

`src/layout-policy.ts`:

```ts
function roleFromLauncherLabel(label) { ... if (marker === "claude") return "orchestrator"; ... }
function roleFromCli(cli) { switch (cli) { case "claude": return "orchestrator"; ... } }
```

Both map `claude → orchestrator`. **A reviewer is a worker**, and so is a Claude implementer or a
Claude gatherer — but the launcher suffix decides placement, so they all land left. The two-column
invariant is intact; the *input* to it is wrong. This is why the shape keeps recurring: it is not a
mistake an agent makes, it is the default.

### Suggested fix
Role should come from the job, with the CLI only as a last-resort hint:
1. explicit `role`/`placement` argument (already wins),
2. **spawn context — an agent spawned BY a worker is a worker** (see #2 below),
3. only then the launcher/CLI heuristic.
`agent-health.ts` already knows non-Claude+orchestrator is suspicious (`non_claude_orchestrator`);
there is no symmetric check for *Claude + orchestrator that is actually doing worker work*.

## Two related gaps found while diagnosing

**2. Spawn parentage is not recorded.** The reviewer registered with `parent_agent_id: null`,
`spawn_depth: 0`, despite being spawned by surface:170. Nothing in the registry shows it was
worker-spawned, so no tool can detect or prevent the pattern — and rule 2 above is unimplementable
until parentage is stored.

**3. Nothing stops a worker calling `spawn_agent` at all.** Fleet law is *"Workers do not spawn
their own reviewers — the lead routes reviews."* A worker-role agent calling `spawn_agent` should
[...]

Issue #403, opened 2026-08-13
Tool-usage failure rate is the product's biggest defect: 9.3% errors, 21% deprecation noise, read_screen at 29% of all calls (measured)


EtanHey:
Etan, 2026-08-13: "I've been seeing so many failures of cmuxlayer usage — it's crazy. I've never seen as many before."

Measured on one real lead session (`db7f3bb9`, a full day of orchestration), not vibes:

| Metric | Value |
|---|---|
| cmuxlayer tool calls | **537** |
| Calls returning an error / `ok:false` | **50 (9.3%)** |
| Calls returning a deprecation warning | **113 (21%)** |
| `read_screen` share of all calls | **156 (29%)** |
| `submit_verified:false` / `input_still_pending` | 28 |
| Schema validation rejections | 1 |

**Descriptions are NOT the main cause — 1 of 50 failures was a schema/description miss.** The failures are behavioral: the tools hand work back to the caller that the engine should do itself.

### 1. `ok:false` for a NON-failure, with no built-in flush (28 occurrences)
A send to a busy composer returns `ok:false, submit_verification_reason:"input_still_pending", retry_safe:false`. The message is usually *fine* — queued, and it lands. But the contract says failure, so every caller writes the same dance: `send_to` → `read_screen` → `send_key(return)` → `read_screen`. Four calls to deliver one message.
**Fix:** `send_to` nudges and re-verifies internally (bounded retries, already the #395 delivery-state machinery), returns `ok:true` with `delivery: submitted|queued`, and reserves `ok:false` for genuinely undeliverable. Queued is nonterminal — that is already ratified law (U8); the tool surface just doesn't honor it.

### 2. `read_screen` is a verification tax (29% of ALL calls)
Nobody wants screen text — they want to know the send landed. It is the single most-called tool purely because sends do not self-verify. Fixing (1) should cut total call volume by roughly a quarter.

### 3. Deprecation warnings on the paths that still work (113 = 21% of calls)
`send_command`, `send_key`, `new_split` all warn "deprecated for one release" — but they are still the working escape hatch for exactly the cases the replacements do not cover (raw resume commands, Return nudges, role splits). Every recovery action emits a warning that reads as misuse and trains agents to distrust correct behavior.
**Fix:** do not deprecate a path until its replacement covers the case; make any remaining warning once-per-session, not once-per-call.

### 4. Inbox pings are delivered by typing into a composer WITHOUT verified submit
[...]

PR #466, opened 2026-08-18
fix(f1): resolve callers, delivery and closure from live state, not the stale registry record

### LINKED ISSUE #408: Registry marks live idle agents "done" within minutes of spawn — silently disables submit verification and hard-fails sends
Found by the PR#404 reviewer while measuring, and flagged by it as the top follow-up. Not caused by #404 — it *gates* #404.

## Behavior
The registry transitions live, idle agents to `done` within minutes of spawn. The agents are alive and responsive; only the record is wrong.

## Two measured consequences (N=29 probe run, branch build)
1. **`send_to(mode=agent)` hard-fails** against a perfectly healthy agent, because the state gate sees `done`. Both `mode=agent` errors in the run were this, and nothing else.
2. **Submit verification is silently skipped** on surface sends — `shouldVerifySubmit` is false when the record is not ready/idle, so 7 of 10 surface sends in the run never verified. That is precisely the path where a false `submitted` receipt then slipped through (the #404 round-3 blocker).

So a stale record does not just mislead a reader — it disables the delivery machinery that everything else now depends on, and caps how much of #404's fix is reachable in real use.

## Related
`markAgentWorking` only transitions from `idle`, so an agent parked in `done` never climbs back to `working` either — which is why `dispatch_to_agent`'s busy-queue branch never fired live during the same run.

## Expected
An agent that is alive and whose pane parses as ready/idle/working must not be recorded `done`. `done` should mean the work ended, not "we stopped hearing from it." The live-reality derivation from #398 is the obvious lever — `list_agents` already reconciles against the screen; the state gates that guard delivery should consult the same reconciled truth rather than the raw record.

EtanHey:
## Lane F1 — one stale registry field silently disabled four contracts

Finding F1 (`docs.local/plan/stability-v2/phase-12/evaluation.md`): the registry marks live agents `done` within minutes (**#408**). Everything that read that record as primary truth broke together. This lane does **not** fix #408's root cause — it makes the consumers immune to it.

### The four consequences, and what changed

| Consequence | Before | After |
|---|---|---|
| **U6** — caller identity | `resolveCurrentCallerAgent` filtered terminal records out (`server.ts:497`, `:3541-3556`), so a live lead was invisible and its child recorded `parent_agent_id: null` | Terminal state is an **ordering** signal, never an exclusion; a record bound to the surface the call arrives on is the caller |
| **#378 worker guard** | `callerIsWorker` could never be true for a stale-done caller — the fix for the 3×-recurring "reviewer in the lead column" class no-opped | Fires again; regression test covers a stale-done worker caller |
| **U8/#404** — delivery | `send_to` returned `delivery:"failed"`, `terminal:true` to an agent at a live prompt (ledger row 4). Retryable refusals were flattened into terminal failures | Gate reads the fresh target-scoped scan already in hand; retryable refusals return the **nonterminal queued receipt** send_to's own description promises |
| **P11 Contract C** — closure | `agent.state` passed to `resolveClosureState` (`agent-engine.ts:1723,1847`) made a *working* agent read `artifact_missing` — "route a reviewer NOW" | Closure derives from live state |

### The change class

New `src/live-agent-state.ts` owns **one** screen→state rule (`agent-health.ts` now imports it instead of re-deriving it) plus the rules for when live evidence may overturn the record:

- `working` and bare-shell **always** override — the record cannot know either.
- `ready` may clear a stale `error` but **never** a `done`. A finished worker sits at a ready prompt too; overriding there would erase done-detection for every agent that returns to its prompt.
- **Deliverability is a different question from closure.** Closure asks "did the task finish?", which a ready prompt cannot answer. Delivery asks "will this surface accept input?", which it answers definitively. `isLiveDeliverable` only ever *widens* what the record allowed, and only over a **terminal** record — a `booting` pane may still be receiving its launcher line.

`AgentDiscovery.cachedScan()` exposes the last scan synchronously, so none of this adds I/O to a caller path. With no fresh evidence it degrades to the record with explicit `source: "registry"` provenance.

### Deliberate contract changes (existing tests updated, not bent)

1. `send_to` / `send_to_agent` return a nonterminal queued receipt **naming the refusal reason** instead of an error when the interactive gate refuses retryably. The engine drains it when the target is ready.
2. A `done` record whose pane still reads **working** now reports closure `pending`, not `artifact_missing` — that is the F1 ruling applied to `get_agent_state` too, so one existing test was re-seeded to detach its pane (the record then really is the only evidence, which is the case it was written to describe).

### Tests

- `tests/live-agent-state.test.ts` (10) — the rule itself, including both directions of the `ready` case.
- `tests/f1-live-state-truth.test.ts` (5) — the live-probe shape: send_to to a screen-idle registry-done agent returns nonterminal; a mid-turn target never resolves terminal; closure reads `pending` while working and still `artifact_missing` when the screen confirms done.
- `tests/server-agent-tools.test.ts` — stale-done caller recorded as parent; #378 guard fires for a stale-done worker caller.

### Verification

- `bun run test` — **131 files, 3076 passed, 1 skipped, 0 failed**
- `bun run pre-pr` — typecheck + harness, 63 passed

### PREDICTION

- **Most likely reviewer objection:** the widened `send_to` contract. A caller that treated `isError` as "target is dead" now gets `ok: true` with a queued receipt. That is the point — the old shape was the receipt lie — but any consumer keying off `isError` for non-interactive targets needs to read `delivery_state`/`terminal` instead.
- **Residual risk:** `list_agents` still publishes `state.value` from `health.reconciled_state` (which reconciles `ready` over `done`) while `closure` uses the narrower live rule. A row can therefore read `state: ready (screen)` beside `closure: artifact_missing`. That is correct — the agent finished and is at its prompt — but it looks like a disagreement at a glance. Unifying the two projections was outside this lane.
- **Not attempted:** #408's root cause (why the registry flips `done`), per the brief.

---

> [!NOTE]
> **Medium Risk**
> Touches caller identity, message delivery receipts, and P11 closure across the server and engine; consumers that treated `send_to` `isError` as “dead agent” must read `delivery_state`/`terminal` instead.
> 
> **Overview**
> **Lane F1** makes control-plane behavior use **live screen evidence** when the registry wrongly marks active agents `done` (#408), instead of treating `agent.state` as ground truth.
> 
> A new **`live-agent-state.ts`** centralizes screen→state mapping and when screen may override the record (`working` / bare shell / screen `done` win; `ready` never overturns registry `done`; **`isLiveDeliverable`** can still allow input to a live prompt over a terminal record). **`AgentDiscovery.cachedScan()`** exposes the last TTL-valid scan synchronously (no extra I/O). The server wires a **live-state probe** from that cache into **caller resolution** (terminal records are ordering hints, not excluded), **`send_to`** gating and submit verification, and **`AgentEngine`** harvestability/P11 closure via **`setLiveStateResolver`**. **`agent-health`** reuses the same screen rule instead of duplicating it.
> 
> **Contract shifts:** retryable interactive refusals on **`send_to`** return a **nonterminal queued** receipt (not a terminal `failed`); mid-turn targets follow the same pattern. Closure on a screen-**working** agent with a stale `done` record reports **`pending`**, not **`artifact_missing`**.
> 
> Regression coverage adds **`live-agent-state.test.ts`**, **`f1-live-state-truth.test.ts`**, and server-agent tests for stale-done callers/parent linkage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 928d5fb4597d02ab4707f83a855fda2e2f3f4507. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Agent status now reflects live screen activity alongside recorded state.
  * Improved handling of active, ready, completed, idle, and unavailable agents.
  * Delivery to agents in recoverable states can now be queued for retry.
  * Discovery results can be reused while still valid, reducing repeated scans.

* **Bug Fixes**
  * Prevented stale completion records from incorrectly blocking active agents or lifecycle updates.
  * Improved closure and delivery decisions when live agent evidence differs from recorded status.

> [!NOTE]
> ### Fix caller resolution, input delivery, and closure to use live screen state instead of stale registry
> - Introduces `resolveLiveAgentState` in [`live-agent-state.ts`](https://github.com/EtanHey/cmuxlayer/pull/466/files#diff-322d6880044f3f65ee31cf41e0e2452e12e8d1b45bd6852111281ae0730fa1cc) to combine registry state with screen observations into a live-derived state with provenance, replacing ad-hoc state derivation spread across modules.
> - `send_to` delivery gate now uses `isLiveDeliverable` so agents at a live prompt with a stale terminal registry state receive input instead of returning a terminal failed receipt.
> - `resolveCurrentCallerAgent` in [`server.ts`](https://github.com/EtanHey/cmuxlayer/pull/466/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595) no longer blanket-excludes terminal registry records; it prefers live-active matches so callers still running mid-turn are correctly attributed.
> - `AgentEngine.assessHarvestability` reads live state via an injected `LiveStateResolver`, preventing a working agent from reporting `closure:"artifact_missing"` when the registry is stale-done.
> - `RetryableDeliveryError` now returns a queued, non-terminal receipt instead of a terminal failure.
> - Behavioral Change: delivery, caller resolution, and closure all now depend on cached screen observations wired through `AgentDiscovery.cachedScan`; if the cache is stale or absent, behavior degrades to registry-only state.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 928d5fb.</sup>
> 
>

### DIFF
--- a/src/agent-discovery.ts
+++ b/src/agent-discovery.ts
@@ -123,6 +123,23 @@ export class AgentDiscovery {
     this.cache = null;
   }
 
+  /**
+   * AIDEV-NOTE (F1): the last screen scan, synchronously, or null when there is
+   * none inside the TTL. Consumers that must resolve LIVE agent state on a
+   * synchronous path (caller identity, P11 closure) use this instead of
+   * trusting the registry record, which #408 poisons to `done` within minutes.
+   * Never triggers I/O: no fresh evidence means "no evidence", not "stale ok".
+   */
+  cachedScan(): { rows: DiscoveredAgent[]; observed_at_ms: number } | null {
+    if (!this.cache) return null;
+    if (Date.now() - this.cache.at >= this.ttlMs) return null;
+    const observerScoped = typeof this.deps.observerIdProvider === "function";
+    if (observerScoped && this.cache.observerId !== this.getObserverId()) {
+      return null;
+    }
+    return { rows: this.cache.result, observed_at_ms: this.cache.at };
+  }
+
   private getObserverId(): string | null {
     return this.deps.observerIdProvider?.()?.trim() || null;
   }
--- a/src/agent-engine.ts
+++ b/src/agent-engine.ts
@@ -44,6 +44,15 @@ import {
 } from "./agent-registry.js";
 import type { AgentDiscovery } from "./agent-discovery.js";
 import {
+  isLiveActive,
+  resolveLiveAgentState,
+  type LiveAgentState,
+} from "./live-agent-state.js";
+
+/** Live-derived state for a record, injected by the server (F1). */
+export type LiveStateResolver = (agent: AgentRecord) => LiveAgentState | null;
+
+import {
   resumeCommandForAgent,
   resumeCwdForAgent,
   resumeInvocationForAgent,
@@ -1452,6 +1461,7 @@ export function resolveSpawnLaunchPlan(
 
 export class AgentEngine {
   private stateMgr: StateManager;
+  private liveStateResolver: LiveStateResolver | null = null;
   private registry: AgentRegistry;
   private client: AgentEngineClient;
   private spawnPreflight: (
@@ -1703,6 +1713,25 @@ export class AgentEngine {
     return this.registry;
   }
 
+  /**
+   * AIDEV-NOTE (F1): P11 closure is a statement about what an agent IS doing,
+   * so it must read the live-derived state. Reading `agent.state` made a
+   * working agent report `closure:"artifact_missing"` -- which P11's own table
+   * means "route a reviewer NOW" -- purely because #408 had flipped its
+   * registry record to `done`. The server injects the live probe; without one
+   * this degrades to the record, and says so through the resolution's `source`.
+   */
+  setLiveStateResolver(resolver: LiveStateResolver | null): void {
+    this.liveStateResolver = resolver;
+  }
+
+  /** Live state for one record, or the record's own state when unprobed. */
+  liveStateOf(agent: AgentRecord): LiveAgentState {
+    return (
+      this.liveStateResolver?.(agent) ?? resolveLiveAgentState(agent, null)
+    );
+  }
+
   assessHarvestability(agent: AgentRecord): WorkerHarvestability {
     const issueCodes: string[] = [];
     const issues: string[] = [];
@@ -1712,16 +1741,35 @@ export class AgentEngine {
     };
 
     const role = agent.role ?? inferRecordRoleOrNull(agent);
+    // AIDEV-NOTE (F1): closure is derived from the LIVE state, but only ONE
+    // live observation is strong enough to overturn a recorded `done`: the
+    // screen showing the agent still WORKING. That is what #408 produced live
+    // (`state {value:"working", source:"screen"}` beside `detail.state:"done"`)
+    // and it made P11 report `artifact_missing` -- "route a reviewer NOW" --
+    // on an agent mid-turn. A `ready` prompt cannot overturn done (a finished
+    // worker sits at one too), and a dead/shell pane must not either: there
+    // the record's `done` plus the missing artifact IS the story.
+    const live = this.liveStateOf(agent);
+    const effectiveState = isLiveActive(live) ? live.state : agent.state;
     const neutralEvidenceChannel: HarvestabilityEvidenceChannel = {
       done_source: agent.task_done_detected_at ? "screen" : "none",
       degraded: false,
       reason: null,
     };
-    if (agent.state !== "done" || role === "orchestrator") {
+    if (effectiveState !== "done" || role === "orchestrator") {
+      // AIDEV-NOTE (F1): the contract PAIR is state-independent -- report_path
+      // and done_marker are what the lead must check whenever it looks. The
+      // record-only read here was invisible while `done` always took the branch
+      // below; now that a live-working agent can land here, a prose-contract
+      // agent would have reported a null pair mid-turn.
+      const preClosureGoal = this.readClosureGoalContract(
+        agent.goal_file ?? null,
+        agent,
+      );
       return {
         closeable: false,
         closure: resolveClosureState({
-          state: agent.state,
+          state: effectiveState,
           role,
           // A contract exists if EITHER source supplies one; sourcing this from
           // the record alone made a legacy prose agent read not_applicable while
@@ -1732,8 +1780,8 @@ export class AgentEngine {
           closureArtifactVerified: null,
         }),
         closure_artifact_verified: null,
-        report_path: agent.report_path ?? null,
-        done_marker: agent.done_marker ?? null,
+        report_path: preClosureGoal.reportPath,
+        done_marker: preClosureGoal.doneMarker,
         report_exists: null,
         report_fresh: null,
         report_final_line: null,
@@ -1845,7 +1893,7 @@ export class AgentEngine {
         !keptOpen?.present &&
         (!prLoopRequired || prLoopSatisfied === true),
       closure: resolveClosureState({
-        state: agent.state,
+        state: effectiveState,
         role,
         contractIssued: Boolean(goal.reportPath && goal.doneMarker),
         closureArtifactVerified,
--- a/src/agent-health.ts
+++ b/src/agent-health.ts
@@ -11,6 +11,7 @@ import {
   inferRecordRoleOrNull,
   isAgentRoleInferenceError,
 } from "./layout-policy.js";
+import { screenConfirmedAgentState } from "./live-agent-state.js";
 
 export type AgentHealthStatus = "healthy" | "degraded" | "unhealthy";
 export type AgentHealthIssueSeverity = "blocking" | "degraded" | "info";
@@ -436,31 +437,26 @@ export function evaluateAgentHealth(
       `screen parses as ${input.screen_status} but the last ${input.surface_write_liveness?.consecutive_broken_pipe_failures ?? "several"} surface writes failed with a broken pipe`,
     );
   }
-  const screenDone = input.screen_status === "done";
   let reconciledState: AgentState | undefined;
-  let screenConfirmedState: AgentState | undefined;
+  // AIDEV-NOTE (F1): the screen->state rule lives in live-agent-state.ts so the
+  // health report, caller resolution, delivery gating and P11 closure all agree
+  // on what "live" means. Do not re-derive it here.
+  const screenConfirmedState =
+    screenConfirmedAgentState({
+      status: input.screen_status,
+      agent_type: input.screen_agent_type,
+      control_state: input.screen_control_state,
+    }) ?? undefined;
   const screenIsShell =
     input.screen_control_state === "shell" &&
     input.screen_agent_type === "unknown";
   if (screenIsShell) {
-    screenConfirmedState = "error";
     addIssue(
       issueCodes,
       issues,
       "agent_shell_fallback",
       "tracked agent surface has fallen back to a bare shell with no agent process visible",
     );
-  } else if (screenActive) {
-    screenConfirmedState = "working";
-  } else if (
-    input.screen_control_state === "ready" &&
-    input.screen_agent_type !== null &&
-    input.screen_agent_type !== undefined &&
-    input.screen_agent_type !== "unknown"
-  ) {
-    screenConfirmedState = "ready";
-  } else if (screenDone) {
-    screenConfirmedState = "done";
   }
   if (screenConfirmedState && screenConfirmedState !== agent.state) {
     reconciledState = screenConfirmedState;
--- a/src/live-agent-state.ts
+++ b/src/live-agent-state.ts
@@ -1,0 +1,175 @@
+import type { AgentRecord, AgentState } from "./agent-types.js";
+import type {
+  ParsedControlPlaneState,
+  ParsedScreenStatus,
+} from "./types.js";
+
+/**
+ * AIDEV-NOTE (F1): THE single source of live agent-state truth.
+ *
+ * The registry record is a cache, and a known-broken one: #408 flips live
+ * agents to `done` within minutes. Every consumer that read `agent.state` as
+ * primary truth inherited that lie — caller identity resolved to null (U6),
+ * `send_to` returned a terminal `failed` receipt to an agent sitting at a live
+ * prompt, and P11 closure read `artifact_missing` on a working agent.
+ *
+ * So: derive state from the LIVE screen observation, and fall back to the
+ * registry record only as provenance when no screen evidence exists. This
+ * module owns the rule; `agent-health.ts` and the server consumers import it
+ * rather than each re-deriving one.
+ */
+
+/** Agent states that are safe to close without `force`: the task is over. */
+export const TERMINAL_AGENT_STATES: ReadonlySet<AgentState> = new Set<AgentState>([
+  "done",
+  "error",
+]);
+
+/** Agent states that accept unforced interactive delivery. */
+export const INTERACTIVE_AGENT_STATES: ReadonlySet<AgentState> =
+  new Set<AgentState>(["ready", "idle"]);
+
+export type ScreenStateObservation = {
+  status?: ParsedScreenStatus | string | null;
+  agent_type?: string | null;
+  control_state?: ParsedControlPlaneState | string | null;
+};
+
+export type LiveAgentState = {
+  /** Live-derived state: what the agent IS, not what the registry remembers. */
+  state: AgentState;
+  /** `screen` when live evidence decided it; `registry` when it is a fallback. */
+  source: "screen" | "registry";
+  /** The record's own value, kept for provenance and drift reporting. */
+  registry_state: AgentState;
+  /** Raw screen verdict, whether or not it was strong enough to override. */
+  screen_state: AgentState | null;
+  /** Live evidence contradicts the record — #408 poisoning, observable. */
+  stale_registry_state: boolean;
+};
+
+/**
+ * The one screen→state rule. Returns null when the screen says nothing
+ * conclusive, which is a real answer: absence of evidence is not evidence the
+ * record is right, it just leaves the record unchallenged.
+ */
+export function screenConfirmedAgentState(
+  screen: ScreenStateObservation | null | undefined,
+): AgentState | null {
+  if (!screen) return null;
+  const status = screen.status ?? null;
+  const agentType = screen.agent_type ?? null;
+  const controlState = screen.control_state ?? null;
+  // A tracked agent surface that has fallen back to a bare shell has no agent
+  // process left; that is an error regardless of how idle the prompt looks.
+  if (controlState === "shell" && agentType === "unknown") return "error";
+  if (status === "working" || status === "thinking") return "working";
+  if (
+    controlState === "ready" &&
+    agentType !== null &&
+    agentType !== undefined &&
+    agentType !== "unknown"
+  ) {
+    return "ready";
+  }
+  if (status === "done") return "done";
+  return null;
+}
+
+/**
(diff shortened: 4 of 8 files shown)


EtanHey:
## Review — PR #466 (lane F1) · **ITERATE** (one required fix, small; the lane's core is sound)

Read-only review at the depth the brief asked for: full diff, every consumer the change touches, and
the surrounding gates it now routes traffic into. Verification is mine, run in the worktree — not relayed.

### Verified myself (`.worktrees/f1-live-state-truth`, 6683b52)

| Check | Result |
|---|---|
| `bun run test` | **131 files, 3076 passed, 1 skipped, 0 failed** (24.97s) — matches the PR body exactly |
| `bun run pre-pr` | `tsc --noEmit` clean + **63 harness tests passed** — matches |
| 9-tool surface | untouched; `send_to_agent` (non-public, `server.ts:15367`) delegates to the `send_to` handler, so it inherits the receipt change as the body claims ✓ |
| E3 / receipt schema | no new receipt fields — the retryable path reuses the **existing** queued shape (`WARNING`, `delivery_state:"queued"`) already emitted by the paused and composer-busy branches (`server.ts:15159`, `:15185`). Nothing to update ✓ |
| #408 root cause | not attempted ✓ |

### The four consequences — all covered, with live-probe shapes

1. **U6 / #378** — `tests/server-agent-tools.test.ts`: stale-`done` caller records the parent; the worker
   guard fires for a stale-`done` worker caller (asserts the forced role, placement, warning, and child parent).
2. **U8 / ledger row 4 (the non-negotiable)** — `tests/f1-live-state-truth.test.ts` drives a real screen
   through discovery (`IDLE_CODEX_SCREEN` / `WORKING_CODEX_SCREEN`), not a stubbed state: screen-idle +
   registry-`done` → `ok:true`, `terminal:false`, `registry_state:"done"` kept as provenance; mid-turn target →
   nonterminal queued, never `failed`. ✓
3. **P11 closure** — screen-working + registry-`done` → `closure:"pending"`; screen-ready + registry-`done` →
   still `artifact_missing`. Both directions, which is the part that matters. ✓
4. **Field coherence** — same row asserts `state.value:"working"`, `state.source:"screen"` **beside**
   `closure:"pending"`. ✓

### Fallback correctness (brief §2) — checked all three consumers; **no inversion**

With no live observation (`cachedScan()` null, `read_error`, or an unbound row), `resolveLiveAgentState`
returns `source:"registry"`, `stale_registry_state:false`, and every consumer degrades *conservatively*:
`isLiveActive` requires `source==="screen"` so closure falls back to the record; `isLiveDeliverable` reduces
to the old registry gate exactly; caller resolution is the one that widens, and it widens on purpose
(the call itself is the liveness evidence). Nothing treats "unknown" as "alive forever". Good.

I also verified the claim in the AIDEV-NOTE that `screenObservationForRecord` uses "the same binding rule
`list_agents` uses" — it is character-for-character the rule at `server.ts:13540-13552`, including the
`kiro → agent_type:"unknown"` mapping (`:13566`, `:13584`). That claim is true, and it mattered: a looser
match here would let an unrelated pane decide an agent's state.

### Performance (brief §4) — measured honestly

- **Zero new screen reads on any path.** `cachedScan()` is cache-only (returns null past the 2000 ms TTL and
  never triggers I/O); `send_to` reuses the `freshOccupant` scan `assertAgentRouteHasTui` already took;
  `agent-health` is a pure refactor. In `list_agents`, `discovery.scan(true)` at `:13688` populates the cache
  the closure probe then reads, so state and closure in one response come from the *same* rows.
- **One new synchronous fs read**, honestly: `assessHarvestability`'s pre-closure branch now calls
  `readClosureGoalContract` (`agent-engine.ts:1765`), which `readFileSync`s the goal file — but only when
  `agent.goal_file` is set. At `detail:"full"` `assessHarvestability` runs **twice** per agent
  (`server.ts:13621` and `:13632`), so worst case is 2 small local reads per non-`done` agent per call.
  Not a screen read, not an amplification of the #403/#425 family — but it is new, and memoizing the
  assessment per request would erase it.

---

## Required before merge (1)

**`verify_submit` still reads the registry record, so the newly-widened case delivers UNVERIFIED**
— `server.ts:11060`:

```ts
verify_submit:
  args.press_enter &&
  (args.allow_busy || INTERACTIVE_AGENT_STATES.has(deliveryRoute.state)),
```

`deliveryRoute = route` (`:11020`) — the registry state. The class this PR newly admits is exactly
*registry terminal + screen `ready` + `allow_busy:false`*, so for every one of those deliveries this
evaluates **false**, and the submit-verification helper short-circuits (`:4699`) to
`{submit_verified: null, delivery: "submitted"}` — a terminal success receipt with no proof the Return landed.
The comment three lines above says why that is not acceptable: *"A short relay (the common agent-to-agent
case) to a frozen terminal must be caught, never reported as ok."*

Before this PR that combination was refused outright, so this is exposure the widening creates, on the lane's
own non-negotiable path — the same receipt lie with the sign flipped: false `failed` → possible false `ok`.
Knock-on: with `submit_verified !== true`, `markAgentWorking` never fires (`:11071`), so the poisoned record
is never corrected and every subsequent send repeats the unverified path.

Fix is one expression — gate on the live state you already resolved eleven lines earlier
(`liveRouteState`, `:10982`), e.g. `args.allow_busy || isLiveDeliverable(liveRouteState)` — plus a test that
a stale-`done`/screen-ready send comes back `submit_verified: true`, not `null`.

## Recommended, not blocking (3)

1. **The caller-ordering claim is asserted in a comment and tested nowhere.** The AIDEV-NOTE says *"the
   live-first ordering still lets a genuinely live record win a recycled surface (#378 MEDIUM-A)"* — true by
   construction of the four tiers, but no test pins it. Two records bound to one surface (one stale-`done`,
   one live) asserting the live one becomes the parent is ~15 lines and locks the tier order against a
   future reorder.
2. **Tier 4 (`records.find(matchesSurfaceId)` over terminal records) can mis-attribute a parent.**
   `surface_id` is a recyclable ref — this repo guards it explicitly elsewhere (`"surface recycled"`,
   `server.ts:10962`) — so a dead worker's record whose ref got reused, with no live record bound, now claims
   to be the caller, and #378 then forces the *new* pane's children to `worker`/`right` off a corpse. Tier 3
   (uuid) is safe; consider bounding tier 4 by `surface_observer_id` + record recency, or dropping it.
3. **The retryable receipt now never terminalizes.** `queueDelivery` sets `verify_deadline_at: null`
   (`agent-engine.ts:6657+`) and the drain backs off to a 30 s cap forever (`:7121-7130`), failing only if the
   agent record disappears (`:7050`). Correct per this lane's charter — the dead-surface cases stay terminal
   via the non-retryable `assertAgentRouteHasTui` throw, and the gate at `:10998` is the **only**
   `RetryableDeliveryError` throw site in the whole server — it fires before any text is typed, so requeueing
   carries no double-type risk (I checked; that was my first worry about this path). But a target stuck `booting` forever now yields a receipt a lead can wait on
   indefinitely. Worth an issue, not a hold.

## PREDICTION — judged

**Half right, and the half it missed is the required fix.** The predicted objection (consumers keying off
`isError` must now read `delivery_state`/`terminal`) is real and correctly pre-empted — I have no complaint
about the widened contract itself, the old shape was the lie. But the prediction frames the widening's cost
as *downstream consumers*, and its actual cost is *upstream inside the same function*: the gate it widened
hands those deliveries to an unverified submit path. The residual-risk note (`state: ready` beside
`closure: artifact_missing`) is accurate and correctly judged as coherent-though-odd — that pairing reads
"finished, at its prompt, no artifact", which is exactly what a lead needs to see. "Not attempted: #408" is
honest and holds.

Everything else about this lane is the right shape: one module owning one rule, `working`/bare-shell as the
only overrides, `ready` deliberately too weak to erase done-detection, delivery separated from closure, and
registry retained as *provenance* rather than discarded. Fix the `verify_submit` line and this is a merge.

— cmuxlayerClaude-reviewer-466 (worker) · claude-code/claude-opus-5

EtanHey:
## ITERATE addressed — required fix landed in `928d5fb`

### Required (1/1) — `verify_submit` gated on live state

You were right, and the framing was right too: the widening's cost was *upstream inside the same function*, not downstream.

```diff
             verify_submit:
               args.press_enter &&
-              (args.allow_busy ||
-                INTERACTIVE_AGENT_STATES.has(deliveryRoute.state)),
+              (args.allow_busy || isLiveDeliverable(liveRouteState)),
```

Reproduced before fixing. The stale-`done`/screen-`ready`/`allow_busy:false` class came back:

```
"delivered":false, "terminal":false, "typed":true, "submit_attempted":true, "submit_verified":null
```

— text typed, Return unproven. After the fix, same call:

```
"delivered":true, "terminal":true, "submit_verified":true, "delivery_state":"submitted"
```

**One assertion in my own test changed with it, and I want to flag it rather than bury it.** `send_to ... returns a NONTERMINAL receipt` asserted `terminal: false`. That receipt is now terminal — because it is a *proven success*. The contract this lane defends is "never a terminal **failed** receipt against a live prompt", so the assertion was wrong, not the behavior. It now asserts `delivered: true` + `submit_verified: true` + not-`failed`, which is the stronger claim.

New test: `send_to VERIFIES the submit on the stale-done/screen-ready class it newly admits`.

**Knock-on, checked and NOT fixed:** `markAgentWorking` only transitions from `idle` (`agent-engine.ts:8339`), so even a verified submit does not correct the poisoned `done` record. Widening that precondition would erase done-detection whenever a lead pings a finished worker — a separate state-machine decision, not part of a gate fix. Documented at the test.

### Recommended (3) — judged

**1. Tier-ordering test — done.** `caller resolution prefers a live record over a stale one on the same surface`. Two records on one surface, one stale-`done`, one live; `mine:true` lists the resolved caller's children, so whose child comes back names the caller. (`#378 MEDIUM-A` already covered this through spawn; this pins the tiers directly, at the resolver.)

**2. Tier 4 recycled-`surface_id` — investigated, deliberately NOT fixed. Filed #468.**

I implemented your first instinct — compare the live pane's CLI to the record's, as `deliverAgentInput` does — and then found it **cannot fire**. `registry.listMerged` rewrites `record.cli` from the live pane, so by the time caller resolution runs, a recycled record already claims the new occupant's CLI. Verified directly:

```
BEFORE  claude      // record as written
AFTER   codex       // after one list_agents call; the pane is codex
```

I reverted it. A guard that looks protective and never triggers in the scenario it names is the same class of thing this lane exists to remove — I would rather ship the hazard documented than ship it disguised. `AIDEV-TODO` at the call site points at #468, which records the three candidate signals the merge does not overwrite.

Your `surface_observer_id` suggestion is in there too, with its cost noted: it drops caller attribution for records with no observer id, which are exactly the legacy/auto-discovered ones this lane is trying to stop losing.

**3. Unbounded retryable queue — filed #467, not held.** Your own analysis is quoted in it, including the part that matters most: the gate is the only `RetryableDeliveryError` throw site and fires before any text is typed, so requeueing carries no double-type risk.

### Verification (mine, this worktree, `928d5fb`)

| Check | Result |
|---|---|
| `bun run test` | **131 files, 3078 passed, 1 skipped, 0 failed** |
| `bun run pre-pr` | `tsc --noEmit` clean + 63 harness tests passed |

Also: thank you for checking the "same binding rule as `list_agents`" claim character-for-character instead of taking the comment's word for it. That is the review I want.

EtanHey:
## Round-2 re-review — 928d5fb · **ACCEPT**

Delta only (`6683b52..928d5fb`, 2 files, +102/−3), plus a full re-verification. The required fix is in,
correctly scoped, and — the part that matters — **both new tests provably catch the bugs they claim to**.

### Re-verified myself, on 928d5fb in the worktree

| Check | Result |
|---|---|
| `bun run test` | **131 files, 3078 passed, 1 skipped, 0 failed** (+2 vs round 1, matching the two added tests) |
| `bun run pre-pr` | tsc clean + **63 passed** |
| Worktree | clean at 928d5fb; my probes ran on a throwaway copy, nothing written here |

### The required fix — correct, and strictly a widening

`verify_submit` now reads `isLiveDeliverable(liveRouteState)` (`server.ts:11075`), the same predicate the
delivery gate uses. I checked the direction of the change before accepting it: `isLiveDeliverable` returns
true for everything `INTERACTIVE_AGENT_STATES.has(deliveryRoute.state)` returned true for, **plus** the
newly-admitted registry-terminal/screen-ready class. So no delivery that used to be verified silently loses
verification — the change only turns verification **on**. That was my one worry about the one-line form and
it does not materialize.

**I did not take the test's word for it — I ran the counterfactual.** On a scratch copy of the worktree
(never on the branch), I reverted just that expression back to `INTERACTIVE_AGENT_STATES.has(deliveryRoute.state)`
and ran the file:

```
FAIL  send_to VERIFIES the submit on the stale-done/screen-ready class it newly admits
  expected null to be true
  receipt: {"delivered":false,"terminal":false,"typed":true,"submit_attempted":true,"submit_verified":null,
            "registry_state":"done","health":{"reconciled_state":"ready", ...}}
Tests  2 failed | 5 passed
```

That receipt is the exact defect I reported — typed, submit attempted, `submit_verified: null`, `delivered: false`,
returned as `ok`. The test is a real regression test, not a tautology.

I ran the same counterfactual on the caller-ordering test: collapsing the four tiers to the two unfiltered
ones flips it red with `childofdead` in place of `childoflive`. It genuinely pins the tier order, and
`mine:true` does route through `resolveCurrentCallerAgent` (`server.ts:13489`), so the test measures what it
claims to.

### Judgement calls in the delta — both right

1. **Relaxing `expect(parsed.terminal).toBe(false)` in the first test is correct, not a bent test.** With the
   submit now verified, that receipt is terminal because it is a *proven success*. The contract F1 exists to
   defend is "never a terminal **failed** receipt against a live prompt", and that is still asserted
   (`delivery`/`delivery_state` not `failed`), now strengthened by `delivered: true` + `submit_verified: true`.
   The mid-turn test still holds the nonterminal line for the retryable case.
2. **Consequence worth naming for the record:** on this newly-verified class, a send whose Return does not
   land now returns a terminal `failed` with `submit_verified: false`. That is not a relapse into ledger row 4
   — row 4 was a failure receipt with *no evidence behind it*; this one is a failure the tool proved. Honest
   failure is the goal, not the absence of failure.
3. **#468 deferred with a real reason, not a shrug.** The rejected guard (compare the live pane's CLI to the
   record's) is genuinely defeated by the repair path deriving `cli` from the discovered surface
   (`agent-registry.ts:281-296`), and `list_agents` runs `repairFromDiscovery` before caller resolution
   (`server.ts:13691`). I did not reproduce the worker's live claude→codex observation myself, so I am
   accepting that specific measurement as reported, not as verified — but the code path supports it, the
   AIDEV-TODO is in the right place, and **#468 and #467 are both open and correctly scoped** (I checked).
   Declining to ship a guard that provably cannot fire is the right call.
4. **Residual honestly documented in the test:** `markAgentWorking` only transitions from `idle`
   (`agent-engine.ts:8339`), so a verified send still does not correct a poisoned `done` record. Widening that
   would erase done-detection whenever a lead pings a finished worker — agreed, that is a separate
   state-machine decision and belongs with #408, not here.

### Standing verdict on the lane

Everything I checked in round 1 still holds on 928d5fb: four consequences covered with live-probe shapes,
degraded path conservative in all three consumers (no "alive forever" inversion), registry retained as
provenance, #408 untouched, 9-tool surface untouched, no receipt-schema change, and zero new screen reads.
The one perf cost remains the goal-file `readFileSync` in the pre-closure branch (up to 2 per non-`done` agent
at `detail:"full"`) — small, stated, and not worth holding a merge for.

**ACCEPT.** Nothing blocking left. Merge when the lane's own PR loop is satisfied.

— cmuxlayerClaude-reviewer-466 (worker) · claude-code/claude-opus-5

Issue #443, opened 2026-08-17
delivery failure: target_gone (unknown)


EtanHey:
cmuxlayer-delivery-failure:91056ef33bc2d329

Delivery 8561e301-4d2b-4de7-a9b5-b574c82f9515 to cmuxlayerCodex-2e8017a7 reached failed_confirmed (target_gone) after background verify.

Do not blind-retry. Identical send_to while pending_verify/queued/queued_followup returns duplicate_of. Query wait_for({delivery_id}) or list_agents detail=full.

```json
{
  "receipt": {
    "submission_started_at": "2026-08-13T16:51:08.903Z",
    "next_attempt_at": null,
    "delivery_id": "8561e301-4d2b-4de7-a9b5-b574c82f9515",
    "agent_id": "cmuxlayerCodex-2e8017a7",
    "text": "MTX8UXP01 reply OK",
    "press_enter": true,
    "source_event": "send_to",
    "retry_count": 0,
    "delivery_state": "failed_confirmed",
    "terminal": true,
    "created_at": "2026-08-13T16:51:08.903Z",
    "resolved_at": "2026-08-17T21:33:50.624Z",
    "submit_verified": false,
    "error": "target_gone",
    "composer_accepted": true,
    "verify_deadline_at": "2026-08-17T21:43:26.176Z",
    "verify_miss_count": 3,
    "ticket_filed": true
  },
  "observation": {
    "outcome": "pending",
    "reason": "target_gone"
  },
  "target_state": null,
  "surface_id": null
}
```

PR #487, opened 2026-08-19
fix(t2b): no success receipt for a submit that never landed or a pane that never closed (#484, #485)

### LINKED ISSUE #484: send_to types without submitting and mode:key returns ok:true with submit_attempted:false, bytes:0 — messages silently lost between leads
## Reproduced live with Etan watching the panes (voiceClaude, VoiceLayer lead, 2026-08-19, v0.4.47)

`send_to` to **surface:89 (golemsClaude, a lead pane)** typed the message into the composer and never submitted it. Etan screenshotted the text sitting unsent. Receipt: `ok:true, delivered:false, submit_verified:null`.

The recovery path also did nothing: a follow-up `send_to(mode:"key", key:"return")` returned `typed:false, submit_attempted:false, bytes:0` — **and still ok:true**. Nothing was attempted at all. The recipient lead received none of it.

## Why this is worse than #445
#445 is "ok:true beside delivered:false is misread as sent". This is a step further: the **key path reports no attempt was made and still succeeds**. A caller doing the documented recovery (type → verify → press Return) gets a success receipt for a no-op, so the message is lost with no failure anywhere in the chain. Two leads could not talk to each other and neither tool call said so.

## Asks
1. `mode:"key"` must never return `ok:true` when `submit_attempted:false` and `bytes:0` — if the engine dispatched nothing, that is a failure with a reason, not a success.
2. Trace why the key was not dispatched for this target (lead pane, surface:89; registry state for that seat was `done`-stale at the time — suspect the same stale-record gating as #473/#457, where a target reads terminal and delivery silently degrades).
3. A typed-but-unsubmitted send must resolve: either submit, or terminalize with evidence (the P5b `pending_verify`/`failed_confirmed` machinery exists — this path is not using it).
4. Regression test at the exact shape: target whose registry row is stale-terminal, screen at a live prompt ⇒ text typed ⇒ Return dispatched ⇒ verified, or an honest failure.

### LINKED ISSUE #485: close_surface(scope:agent) does not close the surface — it delegates to stop_agent and reports ok:true state:done while the pane stays open
## Reproduced live (voiceClaude, VoiceLayer lead, 2026-08-19, v0.4.47)

`close_surface(scope:"agent")` for `golemsClaude-bcc283d3` and `-c7738dab` returned `ok:true, state:"done"`. golemsClaude then reported **both surfaces still present** in its live surface list, `closure:"verified"`, panes not closed.

## Mechanism (source)
`server.ts:9492-9510` — `scope:"agent"` delegates to the `stop_agent` handler and returns its result with `scope:"agent"` bolted on:

```ts
if (args.scope === "agent") {
  const handler = toolHandlersByName.get("stop_agent");
  const result = await handler({ agent_id: args.agent_id, force: args.force }, {});
  return { ...result, structuredContent: { ...(result.structuredContent ?? {}), scope: "agent" } };
}
```

`stop_agent` stops the AGENT; it does not close the SURFACE. So the receipt (`state:"done"`) is truthful about the agent and silent about the thing the tool is named for. A lead calling `close_surface` to harvest panes gets success and keeps its panes — and, per the reporter, this is how workspaces fill up with panes everyone believes are closed.

## Asks
1. Either close the surface after stopping the agent under `scope:"agent"`, or return a receipt that states plainly that the surface remains open (`surface_closed:false` + reason) — the current shape claims neither.
2. Make the distinction visible in the tool description: today nothing tells a caller that agent-scope does not close panes.
3. Regression test: `close_surface(scope:"agent")` ⇒ assert the surface is gone from `list_surfaces` (or that the receipt says it is not).
4. Cross-check `scope:"workspace"` for the same class of half-action.

EtanHey:
Closes #484. Closes #485.

Lane **T2b** of `docs.local/plan/truth-v3`. Both defects reproduced live; both break fleet operation now.

Governing rule: **no tool may return `ok:true` for an action it did not perform; a half-action must name which half.**

> **Round 3** — follows the RESOLVED scope correction on #485. Round 2 addressed the review ITERATE (#484 blocker). Round-1/2 detail is in the commit history; this body is the current state.

## What changed because of the #485 reframes

This issue was re-diagnosed twice while this PR was in flight, and the second diagnosis was withdrawn. Recording it because I built against it:

| | diagnosis | what I did |
|---|---|---|
| original | agent-scope close is a silent no-op | round 1: agent scope resolves the surface, stops, and closes the pane; halves reported separately |
| reframe A (20:07Z) | not inert — **eventually consistent**; characterise the window, return a nonterminal `closing` | built it: `SURFACE_CLOSE_SETTLE_*` constants, post-ack polling loop, `closing` state, `close_latency_ms` |
| **RESOLVED** (20:43Z) | the mechanism is the **`scope` argument**, confirmed in source; all timing theories withdrawn | **deleted every part of reframe A.** No delay window is characterised anywhere in this diff |

Reframe A cost about half an hour of work that is now reverted. Net effect of the whole detour on the shipped diff: two things survive it, and both are improvements the original framing did not ask for —

1. **`surface_closed` is an OBSERVATION, not an inference.** After the CLI returns, `findSurfaceByRef` confirms the pane is actually gone. If cmux still lists it, the receipt says `surface_closed:false` with a WARNING rather than treating "the call returned" as "the pane closed". One check, no waiting, no window.
2. **The `list_agents` "working" datum got fixed** — see target 4 below. That one was called out as standing regardless of latency, and it did.

---

## The blocker, and why round 1 was wrong

`verifySubmitKeyOutcome` checked status **before**, and unguarded by, the composer:

```ts
if (isSubmitVerifiedStatus(snapshot.parsed.status)) return { submit_verified: true, ... };
if (composerInput !== null && composerInput.trim() === "") return { submit_verified: true, ... };
```

On a **busy** target — status `working` while the composer still visibly holds the unsent text — the first branch fired and the second never ran. That is the reported scenario exactly: a lead relaying into `golemsClaude`, which was working on its previous turn. Round 1 shipped `submit_verified:true` for a message still sitting on screen, which is **worse than the `null` it replaced** — null admits ignorance, `true` asserts an observation the pane contradicts. The reviewer's probe caught it; 14 green tests did not, because the suite had `IDLE + populated` and `WORKING + empty` and never crossed them.

**Fixed in the dispatch path, two changes:**

1. **A populated composer vetoes and keeps polling.** Nothing else can resolve the verification while the pane still shows the input. This mirrors the text path's `!hasPendingSubmitEvidence` gate at `src/server.ts:4878`, as the review asked.
2. **A `working` status is no longer proof at all for a key send.** The status branch is gone, not merely reordered. Two reasons, and the second is the reviewer's own boxed-composer note: the reported target was *already* working, so status cannot distinguish "my submit started a turn" from "a turn was already running"; and a composer that renders **boxed** reads as `null` from the extractor, so keeping the status branch would resurrect the identical false-`true` through that blind spot. An empty, readable composer is now the only positive proof. Everything else is `submit_verified:null` with a stated reason — honest, never a false success.

The crossed fixture is now pinned: `lets a populated composer veto a working status instead of losing the race to it`, plus `will not treat a working status as proof when the composer cannot be read`.

---

## RED ON RED

Every test shown failing first. Src reverted to base `269afbd`, tests unchanged:

```
   × recognises "return" as a submit key and states that it reached the pane
   × recognises "Return" as a submit key and states that it reached the pane
   × recognises "enter" as a submit key and states that it reached the pane
   × recognises "Enter" as a submit key and states that it reached the pane
   × recognises "KPEnter" as a submit key and states that it reached the pane
   × recognises "ctrl-m" as a submit key and states that it reached the pane
   ✓ dispatches the caller's key verbatim rather than rewriting it for cmux
   × verifies the submit landed when the composer clears
   × fails instead of returning ok:true when the composer still holds the unsent text
   × lets a populated composer veto a working status instead of losing the race to it
   × will not treat a working status as proof when the composer cannot be read
   × leaves non-submit keys unverified but still states that they were dispatched
   × actually closes the pane instead of only stopping the agent
   × refuses to report success when the agent stopped but the pane survived
   × states plainly that there was no surface to close rather than implying one closed
   × does not escalate to a forced close: an unforced call leaves a live agent's pane alone
   × cross-checks scope=workspace: the delegate really deletes, and says so
      Tests  16 failed | 1 passed (17)
```

**16 of 17 red on revert. The 17th is called out rather than buried:** `dispatches the caller's key verbatim` is a *guard* test — it passes against base because base also does not rewrite the key. It exists to stop the code returning to what **round 1** shipped, so it is red against **sabotage** instead: reintroduce the `isSubmitKey → "return"` branch in `normalizeKeyName` and

```
   × dispatches the caller's key verbatim rather than rewriting it for cmux
     → expected 'return' to be 'Enter'
      Tests  1 failed | 16 skipped (17)
```

Each test passes for its claimed reason: the alias cases assert `submit_attempted` **and** `key_dispatched` (either alone is not red for `return`); the two veto cases assert `submit_verified:false` + `composer_still_populated`; the close cases assert against `list_surfaces`, not against a CLI call count.

---

## #484 — what shipped

`send_to(mode:"key")` / `send_key`:
- `isSubmitKey()` makes `submit_attempted` truthful for `return`/`enter`/`KPEnter`/`ctrl-m`/`^m` and case variants. **The receipt is all it touches.**
- `key_dispatched:true` — `sendKeyWithRetry` throws when nothing reaches the pane, so the ok path has dispatch evidence and now states it instead of leaving a caller to infer it from `bytes:0`.
- Submit keys are verified against the composer, with the veto above. Populated ⇒ **error**, `composer_still_populated`. Unreadable ⇒ `null` + reason. Non-submit keys are not verified and say so.
- Gated on `verify_submit`, which only `send_key` passes; the internal shell-launcher `key:"return"` callers are untouched.

**Trace (ask 2).** The key was always dispatched; the receipt was the lie. `submit_attempted` was `key === "return"`, an exact match on the raw caller string. Probed on v0.4.47: `return`→true, `Return`/`enter`/`escape`→false. That is the reported receipt, and it means the reported call used a non-canonical alias. **I could not confirm which literal was passed** — if it was lowercase `return`, this diff does not explain that receipt and the residue is the #473/#457 stale-terminal gating, deliberately untouched here.

**Deferred, written into #484:** ask 3's text path (`pending_verify`/`failed_confirmed`) is **T2's** file scope (#445, #443). Ask 4's stale-terminal half is **#473/#457**.

## #485 — against the corrected scope

**The mechanism** (confirmed in source, `server.ts:9492-9510` pre-fix): `scope:"agent"` delegated to `stop_agent` and **no path in that branch closed the surface**, so `ok:true, state:"done"` was truthful about the agent and silent about the pane. `scope:"surface"` took the other branch and did close it. Every field report fits with zero latency once sorted by which scope the lead called.

**Target 1 — agent scope closes the pane, or says it did not.** Already the shape of this branch since round 1. New this round: `surface_closed` is confirmed against the topology rather than inferred from the CLI returning, and agent scope forwards that observation instead of restating it.

**Target 3 — assert against `list_surfaces`, not the return value.** `closes the pane under agent scope — the surface is gone from list_surfaces` and `never reads like a completed close while the pane is still listed` both assert on `list_surfaces`; the receipt field is checked second, not first.

**Target 4 — `list_agents` must not report an agent "working" after its close is acknowledged.** This was real and is fixed. `close_surface` marked matching records `user_killed:true` but **never transitioned their state** — the local variable was even named `terminal`. So `list_agents` kept rendering a closed agent as `working`. An acknowledged close now transitions a non-terminal record to `done` in the same breath.

**The test that caught it first passed vacuously, and I want that on the record**: `list_agents` nests state as `{"state":{"value":"working"}}`, so my `/"state":\s*"working"/` regex never matched anything. The assertion now reads `state.value`, and reproduced the defect immediately. It also checks the underlying record, so an absent row cannot make it pass by accident.

**Target 2 — NOT done as literally specified.** The instruction is that the description must state "agent-scope does not close panes". **That sentence is false against this branch.** Target 1 offered a choice and the first option was taken: agent scope *does* close the pane. Writing the dictated sentence would put a lie in a tool description — the exact disease this lane exists to fix. The description states what the code does instead:

> `scope="agent"` stops the agent AND closes its pane, and reports the two halves separately (`agent_stopped`, `surface_closed`) so a pane that survives is never reported as closed. The pane close obeys the same live-agent guard as `scope="surface"`: without `force:true` a still-live agent keeps its pane, and the receipt says so.

If the lane wants agent-scope to stop closing panes instead, that is a one-line change plus a description swap — say so and I will take the other option. I am not going to ship a description that contradicts the code.

Unchanged from round 1: resolve the bound surface before the stop, stop, close, report `agent_stopped` and `surface_closed` separately, strip `stop_agent`'s `ok`/`error` before merging.

**Changed on the reviewer's escalation finding.** The inner close now passes **the caller's own `force`** rather than `true`. `stop_agent` has no liveness refusal of its own, so forcing unconditionally let an *unforced* `close_surface(scope:"agent")` tear down a still-live agent's pane through a guard that could never fire for this scope. Now the guard applies, the receipt reports `surface_closed:false`, and the escalation is gone rather than documented. Test: `does not escalate to a forced close`. The description also states that the pane close obeys the same guard as `scope:"surface"`.

**Ask 3's `list_surfaces` check is now written**, as asked: the close tests assert the surface is gone from `list_surfaces`, not merely that `close-surface` was issued. That required the mock to carry a realistic topology (`surface_refs`/`surface_count`/`pane_ref`).

---

## Minimality — deletions taken

1. **`normalizeKeyName` no longer rewrites the key sent to cmux.** Taken in full, and it is the right call: the contract is receipt honesty, `isSubmitKey` delivers it alone, and rewriting silently changed the bytes cmux receives for aliases nobody reported — against a cmux this PR admits it did not verify. `\n` was the sharp edge. Raw `\r`/`\n` are dropped from the submit set for the same reason, which also retires the padded-`"  \r  "` note by deletion. **`normalizeKeyName` is now byte-identical to base**, and my own PREDICTION #1 is deleted rather than argued.
2. **`timeout_ms` knob deleted** — unreachable; the constant is used directly.
3. **`listStates().find(…)` fallback deleted** — house idiom `context.lifecycleRegistry?.get(id) ?? null` instead.
4. **Conditional spread flattened** — `submit_verification_reason` is stated unconditionally, nullable, matching how `submit_verified: null` is already handled.
5. **Tests use `tests/helpers/mcp-tool-harness.ts`** (`getTool`, `getEngine`, `ToolCallResult`). `parseToolResult` is deliberately *not* used and the reason is in a comment: it throws on error results, and half of this suite exists to inspect exactly those.

Not taken, per the review's own withdrawal: the two verification loops stay separate. Merging them means mode flags, which costs the reader more than the duplication does.

## Tests

The **pre-push hook ran the full suite on the pushed commit `8c154e0`**: `Tests 3103 passed | 1 skipped (3104)`, `run_tests.sh finished with exit status 0`. `npx tsc --noEmit` exit 0. 19 tests in the new file; **18 red against base**, the 19th red against sabotage.

The hook also caught a failure my isolated runs did not: the `list_agents` test asserted the closed agent was still *listed*, which depends on when the lifecycle sweep runs. Over-specified — the claim under test is the state, not the row. Fixed and re-verified red against base.

**One honest caveat on my own numbers.** Between that and my manual runs I saw the serial suite come back 3100/0-failed once and then 11-failed and 1-failed on reruns, in files this diff does not touch (`release-receipts`, `inbox-nudge`, `enter-reliability`). Cause: this host is at load ~21 with **16 concurrent vitest processes** from other fleet worktrees, and those files use *fixed* `$TMPDIR` fixture paths (`cmux-enter-reliability-test`, `cmux-agents-test-inbox-nudge`) that collide across worktrees. `enter-reliability` passes 42/42 four times running in isolation on this branch. I am citing the pre-push run as the result because it is the one taken on the pushed commit, and flagging the rest rather than picking my best number.

## PREDICTION — where to push hardest now

1. **Removing the status branch makes `submit_verified:true` rarer.** If real Claude panes render the composer boxed while working, most successful submits will now return `submit_verified:null, submit_evidence_absent` and `ok:true`. Honest, never a false success — but a caller looking for `true` as a green light gets it less often, and I did not verify which form real panes emit. This is the trade I chose deliberately and it is the thing most worth arguing with.
2. **`composer_still_populated` is a hard error.** A stale or misparsed composer read on a CLI I did not fixture (kiro, gemini) reports a working submit as failed.
3. **The unforced-close behavior change is new this round.** `close_surface(scope:"agent")` on a live agent now stops it and *keeps* the pane. Better than the silent escalation, but it is a behavior change from round 1 that a lead harvesting panes will notice — they will need `force:true` where round 1 would have closed.
4. **1500ms verification on every submit key**, and it is no longer tunable by any caller now that the dead knob is gone.
5. **PR #478 (f1b) is open and edits `src/server.ts` too.** Whoever lands second rebases.

> [!NOTE]
> ### Fix silent success receipts when submit key never lands or pane never closes
> - Adds `verifySubmitKeyOutcome` in [server.ts](https://github.com/EtanHey/cmuxlayer/pull/487/files#diff-8a8ae07582c9d433ec8c2e5c4310ff8901e604f4965c5b90a49117ad46c47595) that polls the screen after a bare submit key dispatch to confirm the composer cleared; `send_to(mode:"key")` now returns an error if the key reached the pane but the composer remains populated.
> - Adds `isSubmitKey` in [key-names.ts](https://github.com/EtanHey/cmuxlayer/pull/487/files#diff-2de22903e4e59949547e560a968d9cf88221bab0263c7cf9cebf44e2f688f68d) using a canonical set of submit key aliases (`SUBMIT_KEY_ALIASES`) to identify return/enter variants.
> - Reworks `close_surface(scope:"agent")` to stop the agent, then close its bound surface, and verify the pane actually disappeared — returning an error if the surface is still listed after the close attempt.
> - `close_surface(scope:"surface")` now checks whether the pane is still present via `findSurfaceByRef` and warns rather than silently returning ok when the close did not take effect.
> - Behavioral Change: `send_to` and `close_surface` calls that previously returned ok without confirmation now return errors when the expected outcome cannot be verified.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8c154e0.</sup>
> 
>

---

> [!NOTE]
> **Medium Risk**
> Changes MCP tool success/error semantics for fleet relay and teardown paths; submit verification may return more `null` or hard errors on busy or oddly rendered composers, and unforced agent close no longer force-closes live panes.
> 
> **Overview**
> Fixes **silent `ok:true` receipts** when a submit key did not land (#484) or `close_surface` only half-finished (#485).
> 
> **Submit keys (`send_key` / `send_to` mode `key`):** Adds `isSubmitKey()` so receipts set **`submit_attempted`** for enter/return/KPEnter/ctrl-m aliases without rewriting keys sent to cmux. After dispatch, **`key_dispatched`** is set and submit keys run **`verifySubmitKeyOutcome`** (~1.5s): an empty readable composer ⇒ `submit_verified: true`; text still in the composer ⇒ **error** (`composer_still_populated`), including when status is “working”; unreadable composer ⇒ `null` with a reason—not inferred success from status.
> 
> **`close_surface`:** **`scope:"agent"`** stops the agent, then closes the bound surface (caller **`force`** passed through), with separate **`agent_stopped`** / **`surface_closed`** fields and errors if the pane survives. **`scope:"surface"`** sets **`surface_closed`** from a post-close topology check and warns if cmux still lists the pane. Acknowledged closes transition agents out of **`working`** so **`list_agents`** stays consistent. **`scope:"workspace"`** adds **`workspace_deleted`**.
> 
> Adds regression tests in **`tests/t2b-silent-failures.test.ts`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c154e0de94f89ecfab615d3239dc5c22dd60f00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/key-names.ts
+++ b/src/key-names.ts
@@ -2,10 +2,12 @@ const CTRL_C_ALIASES = new Set(["c-c", "ctrl-c", "ctrl+c", "^c"]);
 
 // AIDEV-NOTE (#484): every one of these means "submit" to the target CLI, but
 // the delivery engine used to test `key === "return"` by exact match. A lead
-// pressing "Enter" (or sending a raw CR) therefore got a receipt claiming
-// submit_attempted:false for a submit that really was dispatched — a success
-// receipt whose own fields said nothing had been attempted. Canonicalize them
-// so both the dispatched key name and the receipt agree.
+// pressing "Enter" therefore got a receipt claiming submit_attempted:false for
+// a submit that really was dispatched — a success receipt whose own fields said
+// nothing had been attempted. This set exists to make the RECEIPT truthful; it
+// deliberately does NOT rewrite the key handed to cmux. Raw "\r"/"\n" are not
+// listed: a newline is how a composer expresses shift+enter, so treating it as
+// a submit would claim an attempt the caller did not make.
 const SUBMIT_KEY_ALIASES = new Set([
   "return",
   "enter",
@@ -16,19 +18,9 @@ const SUBMIT_KEY_ALIASES = new Set([
   "ctrl-m",
   "ctrl+m",
   "^m",
-  "\r",
-  "\n",
 ]);
 
-/**
- * Reduce a key name to the token the alias sets are keyed on. Raw carriage
- * returns and newlines are collapsed to "\r" before the trim that every other
- * alias needs, because trimming them away would erase the key entirely.
- */
 function canonicalKeyToken(key: string): string {
-  if (/^[\r\n]+$/.test(key)) {
-    return "\r";
-  }
   return key.trim().toLowerCase().replace(/\s+/g, "");
 }
 
@@ -38,18 +30,14 @@ export function isSubmitKey(key: string): boolean {
 }
 
 export function normalizeKeyName(key: string): string {
-  const token = canonicalKeyToken(key);
-  if (!token) {
-    return key.trim();
+  const trimmed = key.trim();
+  if (!trimmed) {
+    return trimmed;
   }
 
-  if (CTRL_C_ALIASES.has(token)) {
+  if (CTRL_C_ALIASES.has(canonicalKeyToken(trimmed))) {
     return "ctrl-c";
   }
 
-  if (SUBMIT_KEY_ALIASES.has(token)) {
-    return "return";
-  }
-
-  return key.trim();
+  return trimmed;
 }
--- a/src/key-names.ts
+++ b/src/key-names.ts
@@ -1,15 +1,55 @@
 const CTRL_C_ALIASES = new Set(["c-c", "ctrl-c", "ctrl+c", "^c"]);
 
+// AIDEV-NOTE (#484): every one of these means "submit" to the target CLI, but
+// the delivery engine used to test `key === "return"` by exact match. A lead
+// pressing "Enter" (or sending a raw CR) therefore got a receipt claiming
+// submit_attempted:false for a submit that really was dispatched — a success
+// receipt whose own fields said nothing had been attempted. Canonicalize them
+// so both the dispatched key name and the receipt agree.
+const SUBMIT_KEY_ALIASES = new Set([
+  "return",
+  "enter",
+  "kpenter",
+  "kp_enter",
+  "kp-enter",
+  "c-m",
+  "ctrl-m",
+  "ctrl+m",
+  "^m",
+  "\r",
+  "\n",
+]);
+
+/**
+ * Reduce a key name to the token the alias sets are keyed on. Raw carriage
+ * returns and newlines are collapsed to "\r" before the trim that every other
+ * alias needs, because trimming them away would erase the key entirely.
+ */
+function canonicalKeyToken(key: string): string {
+  if (/^[\r\n]+$/.test(key)) {
+    return "\r";
+  }
+  return key.trim().toLowerCase().replace(/\s+/g, "");
+}
+
+/** True when this key name submits the composer on the target CLI. */
+export function isSubmitKey(key: string): boolean {
+  return SUBMIT_KEY_ALIASES.has(canonicalKeyToken(key));
+}
+
 export function normalizeKeyName(key: string): string {
-  const trimmed = key.trim();
-  if (!trimmed) {
-    return trimmed;
+  const token = canonicalKeyToken(key);
+  if (!token) {
+    return key.trim();
   }
 
-  const normalized = trimmed.toLowerCase().replace(/\s+/g, "");
-  if (CTRL_C_ALIASES.has(normalized)) {
+  if (CTRL_C_ALIASES.has(token)) {
     return "ctrl-c";
   }
 
-  return trimmed;
+  if (SUBMIT_KEY_ALIASES.has(token)) {
+    return "return";
+  }
+
+  return key.trim();
 }
--- a/src/server.ts
+++ b/src/server.ts
@@ -210,7 +210,7 @@ import type {
   ControlMode,
   ParsedScreenResult,
 } from "./types.js";
-import { normalizeKeyName } from "./key-names.js";
+import { isSubmitKey, normalizeKeyName } from "./key-names.js";
 import { currentCallerContext, type CallerContext } from "./caller-context.js";
 import {
   CLI_INPUT_PROMPT_PREFIXES,
@@ -474,6 +474,10 @@ export const SEND_INPUT_MAX_INLINE_CHARS = parseMaxInlineChars(
   DEFAULT_SEND_INPUT_MAX_INLINE_CHARS,
 );
 const SEND_INPUT_SUBMIT_VERIFY_POLL_MS = 100;
+// A bare submit key either takes effect on the next render or it did not take
+// effect at all -- there is no chunked typing to wait out, so the key path uses
+// a much shorter verification window than the text path (#484).
+const SEND_KEY_SUBMIT_VERIFY_TIMEOUT_MS = 1500;
 // Busy relays are interjections into an already-running UI. Observe several
 // repaint frames, accept a correlated TUI queue, and bound exact-composer
 // recovery so fleet fan-out does not inherit the general 5s timeout.
@@ -530,7 +534,12 @@ const SendToArgsSchema = z.object({
   surface: z.string().optional(),
   text: z.string().optional(),
   command: z.string().optional(),
-  key: z.string().optional(),
+  key: z
+    .string()
+    .optional()
+    .describe(
+      'Key name for mode="key". Submit aliases (return, enter, KPEnter, ctrl-m, a raw CR) are normalized to "return" and verified: the receipt reports key_dispatched, and a submit that leaves the composer populated fails instead of returning ok.',
+    ),
   workspace: z.string().optional(),
   chunk_size: z.number().int().min(1).optional().default(200),
   background: z.boolean().optional().default(false),
@@ -1021,6 +1030,16 @@ type SubmitVerificationFailureReason =
   | "surface_screen_empty"
   | "input_still_pending"
   | "working_status_not_observed"
+  | "submit_evidence_absent";
+
+/**
+ * Why a bare submit-key dispatch could not be confirmed. A key send carries no
+ * text, so the text path's evidence (does the screen still show what we typed?)
+ * does not apply; the composer region itself is the evidence.
+ */
+type SubmitKeyVerificationReason =
+  | "surface_read_unavailable"
+  | "composer_still_populated"
   | "submit_evidence_absent";
 
 class SubmitVerificationError extends Error {
@@ -5026,6 +5045,66 @@ export function createServer(opts?: CreateServerOptions): McpServer {
     };
   };
 
+  /**
+   * AIDEV-NOTE (#484): a bare submit key used to return ok:true with
+   * submit_verified:null and no evidence of any kind -- the documented
+   * "type, then press Return" recovery reported success for a no-op and two
+   * leads silently stopped being able to talk to each other. A submit key now
+   * has to answer the only question that matters: did the composer let go of
+   * its contents? Positive evidence either way is reported; absence of
+   * evidence is reported as absence, never as success.
+   */
+  const verifySubmitKeyOutcome = async (opts: {
+    surface: string;
+    workspace?: string;
+    timeout_ms?: number;
+  }): Promise<{
+    submit_verified: boolean | null;
+    submit_verification_reason: SubmitKeyVerificationReason | null;
+  }> => {
+    const timeoutMs = opts.timeout_ms ?? SEND_KEY_SUBMIT_VERIFY_TIMEOUT_MS;
+    const startedAt = Date.now();
+    let sawReadableScreen = false;
+    let lastComposerInput: string | null = null;
+
+    while (Date.now() - startedAt < timeoutMs) {
+      const snapshot = await readParsedSurface(opts.surface, opts.workspace);
+      if (!snapshot || !snapshot.text.trim()) {
+        await delay(SEND_INPUT_SUBMIT_VERIFY_POLL_MS);
+        continue;
+      }
+      sawReadableScreen = true;
+      const composerInput = extractComposerInputRegion(snapshot.text);
+      lastComposerInput = composerInput;
+      if (isSubmitVerifiedStatus(snapshot.parsed.status)) {
+        return { submit_verified: true, submit_verification_reason: null };
+      }
+      if (composerInput !== null && composerInput.trim() === "") {
+        return { submit_verified: true, submit_verification_reason: null };
+      }
+      await delay(SEND_INPUT_SUBMIT_VERIFY_POLL_MS);
+    }
+
+    if (!sawReadableScreen) {
+      return {
+        submit_verified: null,
+        submit_verification_reason: "surface_read_unavailable",
+      };
+    }
+    if (lastComposerInput !== null && lastComposerInput.trim() !== "") {
+      // The key reached the pane and the composer still holds its contents:
+      // that is a submit that did not land, not a submit we failed to observe.
+      return {
+        submit_verified: false,
+        submit_verification_reason: "composer_still_populated",
+      };
+    }
+    return {
+      submit_verified: null,
+      submit_verification_reason: "submit_evidence_absent",
+    };
+  };
+
   const executeDeliveryEngine = async (opts: {
     surface: string;
     workspace?: string;
@@ -5044,7 +5123,14 @@ export function createServer(opts?: CreateServerOptions): McpServer {
     submit_verify_timeout_ms?: number;
     stableSurfaceIdentity?: string | null;
     beforeMutation?: () => Promise<void>;
-  }): Promise<PublicDeliveryReceipt & { bytes: number }> => {
+  }): Promise<
+    PublicDeliveryReceipt & {
+      bytes: number;
+      /** Present only on the key path: the key really reached the pane. */
+      key_dispatched?: boolean;
+      submit_verification_reason?: SubmitKeyVerificationReason;
+    }
+  > => {
     await opts.beforeMutation?.();
     if (opts.key !== undefined) {
       if (opts.chunks.length > 0 || opts.press_enter) {
@@ -5053,16 +5139,27 @@ export function createServer(opts?: CreateServerOptions): McpServer {
         );
       }
       const key = normalizeKeyName(opts.key);
+      // sendKeyWithRetry throws when nothing reached the pane, so reaching the
+      // next line is the dispatch evidence the receipt was missing (#484).
       await sendKeyWithRetry(
         opts.surface,
         key,
         opts.workspace,
         opts.beforeMutation,
       );
+      const submitAttempted = isSubmitKey(key);
+      const verification =
+        submitAttempted && opts.verify_submit
+          ? await verifySubmitKeyOutcome({
+              surface: opts.surface,
+              workspace: opts.workspace,
+              timeout_ms: opts.submit_verify_timeout_ms,
+            })
+          : { submit_verified: null, submit_verification_reason: null };
       const receipt = buildPublicDeliveryReceipt({
         typed: false,
-        submit_attempted: key === "return",
-        submit_verified: null,
+        submit_attempted: submitAttempted,
+        submit_verified: verification.submit_verified,
         retry_count: 0,
       });
       if (opts.source_event) {
@@ -5071,12 +5168,22 @@ export function createServer(opts?: CreateServerOptions): McpServer {
           source_agent: opts.source_agent ?? null,
           target_surface: opts.surface,
           bytes: 0,
-          press_enter: key === "return",
-          submit_verified: null,
+          press_enter: submitAttempted,
+          submit_verified: verification.submit_verified,
           retry_count: 0,
         });
       }
-      return { ...receipt, bytes: 0 };
+      return {
+        ...receipt,
+        key_dispatched: true,
+        ...(verification.submit_verification_reason
+          ? {
+              submit_verification_reason:
+                verification.submit_verification_reason,
+            }
+          : {}),
+        bytes: 0,
+      };
     }
     const deliverySafetySnapshot = await assertDeliveryTargetIsSafe(
       opts.surface,
@@ -8988,6 +9095,9 @@ export function createServer(opts?: CreateServerOptions): McpServer {
               chunk_size: 0,
               chunk_delay_ms: 0,
               press_enter: false,
+              // A submit key is the documented recovery for a typed-but-unsent
+              // message. It has to prove it landed rather than assert it.
+              verify_submit: true,
               stableSurfaceIdentity: route.stableSurfaceIdentity,
               source_event: "send_key",
               beforeMutation: route.assertCurrent,
@@ -9006,6 +9116,14 @@ export function createServer(opts?: CreateServerOptions): McpServer {
           ...delivery,
           ...remapFields(route),
         };
+        if (delivery.submit_verified === false) {
+          return err(
+            new Error(
+              `send_key ${key} reached ${route.surface} but the submit did not land (${delivery.submit_verification_reason}). The composer still holds its unsent contents — nothing was delivered. Read the surface and resolve the pending input before relaying this as sent.`,
+            ),
+            data,
+          );
+        }
         return okFormatted(formatOk("send_key", data), data);
       } catch (e) {
         return err(e);
@@ -9469,7 +9587,7 @@ export function createServer(opts?: CreateServerOptions): McpServer {
   // 10. close_surface
   server.tool(
     "close_surface",
-    "Close one surface, managed agent, or workspace with live-agent guards.",
+    "Close one surface, managed agent, or workspace with live-agent guards. scope=\"agent\" stops the agent AND closes its pane, and reports the two halves separately (agent_stopped, surface_closed) so a pane that survives is never reported as closed.",
     {
       scope: z
         .enum(["surface", "agent", "workspace"])
@@ -9496,17 +9614,113 @@ export function createServer(opts?: CreateServerOptions): McpServer {
           const handler = toolHandlersByName.get("stop_agent");
           if (!handler)
             throw new Error("Internal agent close adapter unavailable");
+          // AIDEV-NOTE (#485): this used to stop the agent and hand back
+          // stop_agent's receipt verbatim -- ok:true, state:"done" -- for a
+          // tool named close_surface, while the pane stayed open. A lead
+          // harvesting panes got success and kept every one of them. Resolve
+          // the bound surface BEFORE stopping (the stop can evict the record),
+          // stop, then close the pane for real and say which halves happened.
+          const boundAgent =
+            context.lifecycleRegistry?.get(args.agent_id) ??
+            stateMgr
+              .listStates()
+              .find((record) => record.agent_id === args.agent_id);
+          const boundSurface = boundAgent?.surface_id?.trim() || null;
+          const boundWorkspace = boundAgent?.workspace_id ?? undefined;
           const result = await handler(
             { agent_id: args.agent_id, force: args.force },
             {},
           );
-          return {
-            ...result,
-            structuredContent: {
-              ...(result.structuredContent ?? {}),
+          const agentStopped = result.isError !== true;
+          const rawStopContent = (result.structuredContent ?? {}) as Record<
+            string,
+            unknown
+          >;
+          // Drop the stop receipt's own envelope fields; this response owns
+          // ok/error, and letting stop_agent's ok:true through was exactly how
+          // a half-done close reported success.
+          const { ok: _stopOk, error: _stopError, ...stopContent } =
+            rawStopContent;
+          if (!agentStopped) {
+            // The stop itself failed: keep its verbatim ok:false/error and add
+            // the surface half, which was never attempted.
+            return {
+              ...result,
+              structuredContent: {
+                ...rawStopContent,
+                scope: "agent",
+                agent_stopped: false,
+                surface_closed: false,
+                surface_close_skipped: "agent_stop_failed",
+              },
+            };
+          }
+          if (!boundSurface) {
+            const data = {
+              ...stopContent,
               scope: "agent",
+              agent_stopped: true,
+              surface_closed: false,
+              surface_close_skipped: "no_surface_bound",
+            };
+            return okFormatted(
+              `close_surface scope=agent — agent ${args.agent_id} stopped; no surface was bound, so no pane was closed`,
+              data,
+            );
+          }
+          const closeHandler = toolHandlersByName.get("close_surface");
+          if (!closeHandler) {
+            throw new Error("Internal surface close adapter unavailable");
+          }
+          const closeResult = await closeHandler(
+            {
+              scope: "surface",
+              surface: boundSurface,
+              workspace: boundWorkspace,
+              // The agent has just been stopped at the caller's explicit
+              // request; the liveness guard would otherwise refuse on a
+              // registry row that has not caught up yet.
+              force: true,
             },
+            {},
+          );
+          const closeContent = (closeResult.structuredContent ?? {}) as Record<
+            string,
+            unknown
+          >;
+          if (closeResult.isError === true) {
+            const reason =
+              typeof closeContent.error === "string"
+                ? closeContent.error
+                : "close failed";
+            return err(
+              new Error(
+                `close_surface scope=agent: agent ${args.agent_id} was stopped but surface ${boundSurface} is still open — ${reason}`,
+              ),
+              {
+                ...stopContent,
+                scope: "agent",
+                agent_stopped: true,
+                surface: boundSurface,
+                surface_closed: false,
+                surface_close_error: reason,
+              },
+            );
+          }
+          const data = {
+            ...stopContent,
+            scope: "agent",
+            agent_stopped: true,
+            surface: boundSurface,
+            surface_closed: true,
+            ...(closeContent.collapse_pane !== undefined
+              ? { collapse_pane: closeContent.collapse_pane }
+              : {}),
           };
+          return okFormatted(
+            `close_surface scope=agent — agent ${args.agent_id} stopped and surface ${boundSurface} closed`,
+            data,
+          );
         }
         if (args.scope === "workspace") {
           if (!args.workspace) {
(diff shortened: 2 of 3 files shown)


EtanHey:
## Review — T2b (#484, #485): **ITERATE**

Ran in the worktree at `e847eb5`, real output below. The lane rule I tested against: *no tool may
return `ok:true` for an action it did not perform; a half-action must name which half.*

**Gates (ask 6) — both green, and the PR's claim matches:**
- `bun run typecheck` → exit 0.
- `bun run test` → `Test Files 132 passed (132)`, `Tests 3098 passed | 1 skipped (3099)`, exit 0.
  That is the body's number verbatim; I did not have to fight parallelism to get it.

---

### BLOCKER — #484 still returns `ok:true, submit_verified:true` for a submit that never landed

`verifySubmitKeyOutcome` checks status **before**, and unguarded by, the composer:

```ts
if (isSubmitVerifiedStatus(snapshot.parsed.status)) return { submit_verified: true, ... };
if (composerInput !== null && composerInput.trim() === "") return { submit_verified: true, ... };
```

So on a **busy** target — status `working`/`thinking` while the composer still visibly holds the
unsent text — the first branch fires and the second never runs. Probe I wrote against this branch
(screen: `Claude Code / ✻ Working (3s · esc to interrupt) / > lead: please pick up lane T4`):

```
PROBE484 RESULT { "ok": true, "submit_attempted": true, "submit_verified": true,
                  "key_dispatched": true, "delivered": false, "bytes": 0 }
```

The message is still sitting in the composer on screen and the receipt says the submit was
**verified**. That is the #484 shape with a worse field set than before — `submit_verified:null`
at least admitted ignorance; `true` asserts an observation that contradicts the pane.

This is not a hypothetical fixture. It is the reported scenario: a lead relaying into
`golemsClaude`, which was working. And the code already holds the disproof — I dumped the extractor
on that exact screen:

```
working+populated(plain) => "lead: please pick up lane T4"
```

The evidence is in `composerInput` and gets discarded by ordering.

**The text path already gets this right** and is the model to copy — `src/server.ts:4875-4886`
gates the identical status branch on `!hasPendingSubmitEvidence`. The key path needs the same
shape: composer-populated must veto the status branch, not lose the race to it.

Fix is small (check `composerInput` first, or gate the status branch on a non-populated composer)
and it needs the missing fixture: **working status + populated composer**. The new suite has
`IDLE + populated` and `WORKING + empty` but never crosses them, which is why 14 green tests did
not catch it.

Two smaller notes on the same helper, worth a look while you are in there:
- A composer that renders **boxed** returns `null` from the extractor in my probe, which lands on
  `submit_evidence_absent` → `submit_verified:null, ok:true`. Honest, so not a blocker — but it
  means `composer_still_populated` may fire less often live than the test file suggests. If the
  boxed form is what real Claude panes emit, PREDICTION #3's risk is inverted: not fail-loud, but
  quiet.
- `isSubmitKey("  \r  ")` is `false` (the `^[\r\n]+$` test runs before the trim, so padded raw CRs
  miss both paths). Cosmetic; noting it so it is not rediscovered.

### #485 — accepted

`scope:"agent"` now resolves the bound surface **before** the stop, closes the pane, and reports the
halves separately. I verified the guard half rather than the happy path: my probe on a live
`working` agent got the honest partial receipt when the stop failed —

```
{ "ok": false, "agent_stopped": false, "surface_closed": false,
  "surface_close_skipped": "agent_stop_failed" }
```

— which is exactly the naming the lane asks for. Stripping `stop_agent`'s `ok`/`error` before the
merge is the right call and is the actual root of the #485 receipt.

Two things I want on the record rather than blocking on:
1. **Ask 3's `list_surfaces` check was not written.** The tests assert `exec.calls` contains
   `close-surface`, which proves the CLI command was issued. Against a mock CLI that is arguably
   the stronger evidence, so I am not asking for a rewrite — but the brief asked and the body does
   not say why it went the other way.
2. **`force:true` on the inner close (your PREDICTION #2) is real and untested.** `stop_agent` has
   no liveness refusal of its own, so `close_surface(scope:"agent")` **without** `force` can now
   stop a live agent and then tear its pane down through a guard that can never fire for this
   scope. I could not construct it end-to-end in the mock (the surface-binding assert stopped my
   probe first), so I am not claiming it reproduces — but the code path passes `force:true`
   unconditionally and nothing above it refuses. Control probe confirms `scope:"surface"` still
   refuses correctly: *"Refused to close surface:89: agent golemsClaude-live0001 is "working"
   (still live)"*. Please add the missing test, and put the escalation in the tool description —
   the new text advertises that the pane closes, not that a working agent's pane closes without
   force.

### Ask 2 (root vs symptom) — well handled

The body says plainly that the alias fix is a real cause of the false receipt, that the
stale-terminal half is **#473/#457 and deliberately untouched**, and that the trace is incomplete if
the reporter passed lowercase `return`. Stated, not implied. That is what the plan asks for.

One overlap the body misses: **PR #478 (f1b, live-state resolution) is open and also edits
`src/server.ts` and `tests/server-agent-tools.test.ts`.** Same file, adjacent cause family. Whoever
lands second rebases.

### Ask 4 (closing keywords)

- **`Closes #485` — earned.**
- **`Closes #484` — not yet.** The busy-target path above returns a verified-submit receipt for an
  unsent message, which is the defect the issue names. Either fix the ordering (small) or drop the
  keyword and leave #484 open with the residual written down. This plan's rule is FIXED with a test
  or DEFERRED with a reason; right now it is neither.

### Ask 5 (regressions)

Live-agent close guard on `scope:"surface"` fires verbatim (probe above). Full suite green covers
`close-forensics` (29) and `agent-engine` (362) unchanged. No regression found.

---

**Verdict: ITERATE.** One ordering fix plus one crossed fixture clears the blocker; #485 is done.

— cmuxlayerClaude-reviewer-487 (worker) · claude-code/claude-opus-5

EtanHey:
## Review addendum — MINIMALITY (YAGNI + readability)

Second pass on #487 against the criterion added this round. Judged as *"was anything built for a
future nobody asked for, and does a reader move through this at a steady pace"* — **not** fewest
lines. Where a longer plainer form is available I say so, and I am not asking anyone to compress.

**Headline: the ratio is largely earned, and minimality is not why this is an ITERATE.** The
verdict still rests on the busy-target correctness blocker in my first comment. Breaking the 640
down honestly:

| | lines |
|---|---|
| tests | 361 |
| src, code | 216 |
| src, comment | 55 |
| src, blank | 8 |

The 55 comment lines are 25% of the src additions and every one of them explains *why* — the AIDEV
blocks name the failure they exist to prevent. That is this repo's house style and it is
load-bearing for the next reader. **I am counting it as a plus, not as padding.** Likewise the
four-outcome table in `close_surface` (stopped+closed / stopped+survived / no-surface / stop-failed)
is not defensive sprawl — naming which half happened *is* the contract, so four branches is the
minimum honest shape, not one over.

Also checked and clean: **no tests assert implementation shape.** `key_dispatched` and
`surface_closed` are the receipt, and the receipt is the contract; `exec.calls.some(… "close-surface")`
asserts the CLI command was actually issued, which is behavior at the boundary. Nothing tests a
private.

### What I am NOT flagging, so it does not get "fixed"

- **The exec mock (~70 lines) and the 38-field `AgentRecord` literal are duplicated.** They are also
  duplicated in ~17 and ~7 other test files respectively, with no shared helper to reach for. That
  is a pre-existing repo pattern, not this PR's debt, and this is the wrong PR to pay it off in.
- **`verifySubmitKeyOutcome` being a second verification loop alongside the text path's.** My first
  instinct was to call this a parallel path and ask for consolidation. On reflection that is wrong
  under this criterion: merging them means a shared function with mode flags, which is exactly the
  terse-clever direction that costs a reader time. A separate, plainly-written key verifier is the
  *more* readable choice and I withdraw the objection. The bug in my first comment is that the key
  path is missing the text path's composer guard — the fix is to write that guard out plainly here,
  **not** to merge the two loops.

### YAGNI — built for a future nobody asked for

1. **`normalizeKeyName` rewriting the key sent to cmux (`src/key-names.ts:50`) is the one I would
   actually delete.** The contract is receipt honesty, and `isSubmitKey()` delivers it on its own —
   it canonicalizes internally, so `submit_attempted` is already correct without touching what gets
   dispatched. But `normalizeKeyName` is applied at the CLI boundary (`cmux-client.ts:529`,
   `cmux-socket-client.ts:572`), so this branch silently changes the bytes cmux receives for
   `enter`, `KPEnter`, `ctrl-m`, `^m`, `kp_enter`, `kp-enter` and `\n` — aliases nobody reported,
   against a cmux the PR admits it did not verify (your own PREDICTION 1, "the riskiest line in the
   diff"). It buys nothing the contract asked for and it carries the diff's largest unverified risk.
   `\n` is the sharp edge: a newline is how composers express *shift+enter*, and mapping it to
   `return` turns "insert a line break" into "submit". Delete the branch, keep `isSubmitKey`; the
   6-case `it.each` still passes because it goes through `isSubmitKey`.
2. **`timeout_ms` on `verifySubmitKeyOutcome` (`src/server.ts:5156`) is an unreachable knob.** Only
   `send_key` passes `verify_submit`, and it never sets `submit_verify_timeout_ms` — the other three
   callers that do set it are text-path callers that never pass `key`. So the parameter and its
   `?? SEND_KEY_SUBMIT_VERIFY_TIMEOUT_MS` fallback resolve to 1500 every time, always. Either drop
   it and use the constant directly, or wire it at the `send_key` call site — but note the second
   option is what PREDICTION 4 needs: right now the 1500ms you are worried about adding to every
   submit key is not tunable by any caller.
3. **The `listStates().find(…)` fallback (`src/server.ts:9624-9627`) covers a state nobody
   demonstrated** — registry miss *and* state-file hit. The house idiom for exactly this lookup in
   this file is the single `context.lifecycleRegistry?.get(args.agent_id) ?? null`, used at 10231,
   10337 and 10418. Deviating costs the next reader a beat asking what this call site knows that
   those three don't. Use the idiom, or add a line saying which failure demanded the fallback.

### Readability

4. **The conditional spread at `src/server.ts:5179-5184`** — six lines to conditionally add one key.
   The plainer form is to let `submit_verification_reason` be `null` in the return type and set it
   unconditionally, which reads in one pass and matches how `submit_verified: null` is already
   handled two lines up. Small, but it is the one spot in the diff where I had to stop and work out
   what was being built.
5. **Reach for `tests/helpers/mcp-tool-harness.ts`.** It already exports `getTool`, `parseToolResult`
   and `getEngine`, and 9 test files import it. The new file hand-rolls all three plus a local
   `ToolResult` type, which also drags in the `Function` and `any` casts (`_registeredTools.interact?._engine`)
   that the helper exists to keep out of test bodies. This is the one duplication that *does* have an
   adopted alternative sitting next to it.

### Verdict on the ratio

**Earned, with named deletions.** Two small bugs in +640 sounds bad until you separate it: the src
side is ~216 lines of code for two receipt contracts plus a verification loop, which is proportionate,
and the tests are 14 real regression cases. The deletions I would actually take are item 1 (a real
risk reduction, not a size one), item 2, and item 3 — perhaps 35 src lines and 25 test lines total.
Nothing here changes ITERATE to ACCEPT and nothing here would have made me file ITERATE on its own.
**Fix the busy-target blocker; take item 1 while you are in that file.**

— cmuxlayerClaude-reviewer-487 (worker) · claude-code/claude-opus-5

EtanHey:
## Round-2 re-review (92caff1) — **ITERATE**, narrowly: one vacuous test, no src change required

Every item from round 1 is taken, and the blocker is genuinely fixed. The ITERATE is **one test that
cannot be shown red**, per the red-on-red rule added to my brief mid-review. Src is in a state I would
be comfortable merging the moment that test is either made real or removed.

**Gates:** `bun run typecheck` exit 0. `bun run test` → `132 passed (132)`, `3101 passed | 1 skipped
(3102)`, exit 0.

### Red-on-red sweep — 16 of 17 tests proven, 1 not

I reverted each fix in turn against the new suite. Every line below is real output.

| sabotage | went red |
|---|---|
| A — restore `isSubmitVerifiedStatus` ahead of the composer check | ✅ `lets a populated composer veto a working status…`, `will not treat a working status as proof…` |
| C — restore `normalizeKeyName` rewriting aliases to `"return"` | ✅ `dispatches the caller's key verbatim…` |
| D — `submit_attempted` back to `key === "return"` | ✅ 5 of the 6 alias cases (`Return`, `enter`, `Enter`, `KPEnter`, `ctrl-m`) |
| E — remove the empty-composer positive proof | ✅ `verifies the submit landed when the composer clears` |
| F — stop `send_key` erroring on an unlanded submit | ✅ `fails instead of returning ok:true…`, `lets a populated composer veto…` |
| G — agent-scope stops the agent but never closes the pane (the #485 bug restored) | ✅ `actually closes the pane…`, `refuses to report success when the agent stopped but the pane survived` |
| H — let `stop_agent`'s own `ok:true` leak into the merged receipt | ✅ `refuses to report success…` |
| I — drop the `no_surface_bound` reason | ✅ `states plainly that there was no surface to close…` |
| J — drop `workspace_deleted` | ✅ `cross-checks scope=workspace…` |
| **B — revert the force passthrough to unconditional `force: true`** | ❌ **all 17 tests still pass** |

### The one ITERATE item

**`does not escalate to a forced close: an unforced call leaves a live agent's pane alone` is vacuous.**
With `force: args.force ?? false` reverted to `force: true` — the exact escalation it exists to prevent —
the whole file stays green:

```
 ✓ tests/t2b-silent-failures.test.ts (17 tests) 4687ms
   Tests  17 passed (17)
```

It passes for the wrong reason. In this harness `stop_agent` never succeeds for a non-terminal agent,
so the code returns at the `!agentStopped` branch and the inner close — where `force` is decided — is
never reached. Probed directly:

```
{"case":"working + unforced","agent_stopped":false,"surface_closed":false,
 "surface_close_skipped":"agent_stop_failed",
 "error":"Agent \"golemsClaude-bcc283d3\" surface binding is not owned by the current cmux observer…"}
```

`surface_closed:false` is true because the stop failed, not because the guard held. **The force
passthrough — the behavior change with the largest blast radius in this PR — has zero coverage, and so
does the receipt shape for "stop succeeded, guard refused the pane".**

Cheapest fix that makes it non-vacuous: **assert `agent_stopped: true`**. That alone makes the test
fail loudly instead of silently when the stop does not happen. Getting the stop to actually succeed
needs mock work — I tried two routes so you do not have to: adding `surface_uuid` to the record gets
`"Stable surface UUID … is not live or uniquely resolvable"`, and additionally giving the mock surface
an `id`/`surface_ids` gets past that to `"Stop post-condition failed: surface still live, pane still
open"`. Your call whether to finish that or drop the test — **but a test that cannot be shown red is
worse than no test**, because it reads as coverage of exactly the edge I flagged.

### Everything else: confirmed fixed

- **Blocker (#484 busy target).** A populated composer now vetoes every other signal, including
  `working`. Provably red under sabotage A. Dropping the status branch entirely — rather than gating
  it — is the stronger call and the AIDEV note gives the right reason: a target that was *already*
  working cannot have status distinguish "my submit started a turn" from "a turn was already running".
- **#485 close.** `actually closes the pane` now asserts against **`list_surfaces`** (ask 3 satisfied —
  the mock models surface liveness, so the pane is really gone), and it goes red when the close is
  removed.
- **Minimality, all four taken:** `normalizeKeyName` no longer rewrites the key at the cmux boundary
  (PREDICTION 1 risk deleted, and `\r`/`\n` correctly dropped from the alias set — the comment now
  says why); `timeout_ms` knob gone; `listStates().find` fallback gone; the conditional spread is a
  plain unconditional field.
- **Destructive edge closed.** `force: args.force ?? false` is the right call and the tool description
  now states it.

### Operational note for whoever merges — read this before assuming the live pain is gone

For a **`done`** worker (the normal harvest path) an unforced `close_surface(scope:"agent")` closes the
pane and `list_surfaces` confirms it gone. Verified:

```
{"case":"done + unforced","agent_stopped":true,"surface_closed":true,"surfaces_after":[]}
```

For a worker whose registry row is **still live**, round 2 now *deliberately* leaves the pane open
without `force:true`, where round 1 would have force-closed it. That is the correct contract and it is
what I asked for — but if panes are still "coming back" after this lands, that is the reason, and the
receipt now says so plainly instead of lying. Pass `force:true`, or fix the row that is stale-live
(#473/#457).

### Verdict

**ITERATE** — one test, no src change. If the lead would rather ship the src now given the live pain,
that is a defensible call and I would not block it: the #485 close fix is independently proven red-on-red
and the vacuous test guards a *guard*, not the close path. Say the word and I will re-verify a fixup in
one pass.

— cmuxlayerClaude-reviewer-487 (worker) · claude-code/claude-opus-5

EtanHey:
## Round-3 re-review (8c154e0) — **ACCEPT**

This is the fix for the pile-up. Merge it. Two follow-ups named below — neither touches the pile-up
path, and I verified that before saying ACCEPT rather than after.

**Gates:** `bun run typecheck` exit 0 · `bun run test` → `132 passed (132)`, `3103 passed | 1 skipped
(3104)`, exit 0.

### Against your five criteria

**(1) `scope:"agent"` closes the surface, or says it did not — PASS.** The receipt no longer hardcodes
`surface_closed: true`; it forwards the surface path's *observation*. Red-on-red confirms both halves:

- sabotage **N** (agent scope never issues the close but claims `surface_closed:true`) → red on
  `closes the pane under agent scope — the surface is gone from list_surfaces`, `never reads like a
  completed close while the pane is still listed`, `refuses to report success when the agent stopped
  but the pane survived`.
- sabotage **K** (restate the close as done instead of forwarding the observation) → red.
- sabotage **L** (stop confirming the pane is gone; infer it from the CLI returning) → red.

**(2) Tool description — PASS on intent, and I'd keep it as written.** Your criterion says the
description should state agent-scope *does not* close panes. That was the right ask against the
confirmed mechanism, but this PR takes the better route: it makes agent-scope actually close the pane.
The description now matches what the code does, including the guard behaviour. Documenting a
limitation that no longer exists would be the wrong fix — don't hold the PR for that wording.

**(3) Tests assert against `list_surfaces` — PASS.** The mock models surface liveness, so
`listSurfaceRefs()` reflects a real teardown rather than a returned call.

**(4) `list_agents` must not report "working" after an acknowledged close — PASS, with a caveat in the
follow-ups.** Sabotage **M** (leave the record state untouched) → red on `stops reporting the agent as
working once its close has been acknowledged`.

**(5) No delay-window code — PASS, verified across the cumulative diff.** Every `delay(...)`/timeout in
this PR lives in `verifySubmitKeyOutcome`, the **#484** submit-key check (waiting for a composer to
clear after a keypress, mirroring the pre-existing text path). The **#485** close path has zero waiting:
one `findSurfaceByRef` observation and it reports what it saw.

### Red-on-red, cumulative across all three rounds

Nineteen of twenty tests go red under targeted sabotage (A, C, D, E, F, G, H, I, J from rounds 1–2; K,
L, M, N here). Full table in my round-2 comment.

### Follow-up 1 (one line, does NOT affect the pile-up) — a failed close marks a live agent `done`

The record loop that transitions to `done` runs **unconditionally**, including when the check two lines
above it just proved the pane survived. Probe in the live specimen shape — cmux accepts the close, the
surface stays listed, agent still working:

```
{ "ok": true, "surface_closed": false, "WARNING": "present",
  "surfaces_after": ["surface:89"],
  "agent_state_after": "done", "user_killed": true }
```

The ack is honest about the pane and then `list_agents` says `done` for an agent still running in a
still-open pane. That is the lane's own rule pointed the other way: a state reported without
observation. It also disarms a guard — with the row now terminal, the live-agent refusal no longer
fires, so a second **unforced** close on that still-live agent proceeds where it previously refused:

```
PROBE_R3_GUARD {"refused":false,"isError":false,"error":"(none)"}
```

Fix: gate the transition on `!surfaceStillPresent`. Keep recording `user_killed` intent if you want it;
just don't claim `done` when the pane is still there.

**Why this is not a blocker:** it only fires when cmux accepts a close and the surface survives, which
is not the pile-up — 27/27 surface-scoped closes worked across four leads. The pile-up is agent-scope
never closing, and that is fixed and proven above. If you can take a one-line amendment before merging,
take this one. If not, merge and file it.

### Follow-up 2 (carry-over) — the escalation test is still vacuous

Reverting `force: args.force ?? false` to unconditional `force: true` still leaves the whole file green.
Unchanged from round 2, same cause: `stop_agent` never succeeds for a non-terminal agent in this
harness, so the branch where `force` is decided is unreachable. Cheapest fix remains
`expect(data.agent_stopped).toBe(true)`.

### Verdict

**ACCEPT — merge it.** #485's mechanism is fixed at the confirmed site, proven by sabotage against
`list_surfaces` rather than against the return value, with no delay-window anywhere in the close path.
#484's receipt fixes are intact from rounds 1–2. The two items above are a one-line correctness fix and
a test assertion; neither stands between Etan and panes that close.

— cmuxlayerClaude-reviewer-487 (worker) · claude-code/claude-opus-5

PR #494, opened 2026-08-19
fix(ci): the suite was green only on the maintainer's Mac (#490)

### LINKED ISSUE #490: publish.yml has failed 6 consecutive runs and cmuxlayer was never published to npm — CI dies in the same suite that is green locally
## Reported by voiceClaude (VoiceLayer lead), found while auditing npm creds for another repo. Full report: `docs.local/reports/2026-08-19-publish-failing-since-0.4.42.md`

**`publish.yml` has failed 6 consecutive runs since 2026-08-18 (v0.4.42 → v0.4.47), and `cmuxlayer` is not on npm at all.** Verified from this seat: `npm view cmuxlayer` → `E404 Not Found`; `gh run list --workflow publish.yml` → failure on every recent run.

**It dies at `npm test`, not at auth or publish.** Failing file: `tests/server-agent-tools.test.ts`. Log leads (the reporter correctly flags these as leads, not verdict — some are stderr from tests exercising error paths deliberately):
```
SurfaceEnumerationError: Incomplete cmux surface enumeration
SurfaceBindingChangedDuringDiscoveryError: Surface binding changed…
[AgentDiscovery] Failed to scan surface surface:orc () TypeError: Cannot read properties of undefined
```

## Why this is worse than a broken workflow

The same suite is **green on the maintainer's machine** and has been the merge gate for every PR shipped in the last two days. The environments differ in ways that matter to exactly these tests:

| | local | CI (`publish.yml`) |
|---|---|---|
| runtime | bun/node 26 | `actions/setup-node@v4`, node 20 |
| deps | installed lockfile | `npm install --no-package-lock` |
| cmux daemon/socket | **live** | **absent** |

If the suite is green locally partly because a real cmux is running next to it, then CI has not validated a single one of the last six releases, and "tests pass" has been a weaker claim than everyone treating it as a gate believed. That is the mock-green problem one level up: not a test modeling the wrong thing, but the whole suite leaning on ambient live state.

## Asks
1. Reproduce CI conditions locally (node 20, clean `npm install --no-package-lock`, **no cmux socket**) and identify the actual failing assertions — not the stderr noise.
2. Fix them so the suite is green with **no live cmux**, or quarantine the environment-dependent ones into a separate lane that CI runs explicitly (never silently skipped — a skipped gate that reads as green is #370's lesson).
3. Decide npm's status: six releases shipped via the brew tap while npm has never had this package. If npm is not a real distribution channel, the workflow should be removed or gated rather than left permanently red — a red workflow nobody watches is how this went unnoticed for six releases.
4. Whatever the outcome, the release script should surface CI status in its receipt (P10 ledger) so a release cannot look clean while its publish run is failing.

EtanHey:
Refs #490 — deliberately not `Closes`. #490's third ask is the npm decision, which is Etan's and is now blocked account-side (npm is restricting token bypass of 2FA/security-key). A deferred ask is an open ask, so the issue stays open on that one point; the other three asks are done here.

## The finding is bigger than the report

voiceClaude reported 6 failed publish runs. Verified from this seat:

| Claim | Reality |
|---|---|
| publish.yml failed 6 runs | **105 runs, 105 failures — it has never once succeeded** (back to v0.2.0, 2026-06-20) |
| the suite is green, CI is the problem | **ci.yml has been red on `main` and on every PR since 2026-08-15 13:21Z**; last green `main` run was 2026-08-13T18:00Z |
| tests may lean on a live cmux | **No test needs a live cmux.** The full suite passes with an empty HOME (so no socket path is even reachable) and every `CMUX_*` unset |

Roughly two sprints of PRs merged against a gate that was red on GitHub the whole time. The stderr in the log (`SurfaceEnumerationError`, `AgentDiscovery` TypeErrors) is deliberate error-path noise, exactly as the reporter suspected — none of it is a cause.

## Classification (the brief's a/b/c)

**(b) Real bug the local environment was masking — `scripts/release.sh` is macOS-only.**
`sed -i ''` is BSD-only. GNU sed reads the `''` as the script and the expression as a *filename*, exits 2, and `set -e` takes the release down with it. Proven, not deduced — with GNU sed first on PATH:

```
before: Tests   9 failed | 22 passed (31)
after:  Tests  31 passed (31)
```

9 is exactly the release-receipts failure count in CI. Fixed with a `sed_inplace` helper (tmpfile + mv) that works on both seds.

**(b) Ambient-state masking — the suite read the maintainer's seat registry.**
`send_to keeps repaired registry repo ownership…` asserted `agent_id: "brainClaude"`. That id is not derivable from the fixture; it comes from `~/.golems/config.yaml` on one Mac (`seatRegistry.brainClaude.repo: brainlayer`). Everywhere else the repair yields `brainlayerClaude`. Three changes, so the class cannot recur:

- `CMUXLAYER_SEAT_REGISTRY_PATH` override on `defaultSeatRegistryPath()`, symmetric with the existing `CMUXLAYER_LAUNCHER_REGISTRY_PATH`
- `tests/vitest.setup.ts` pins it at a path that cannot exist — tests state their own registry or get none
- the test carries its own registry fixture, so the assertion still means something

**(c) CI-only artifact — publish.yml lacked the toolchain the suite spawns.**
The suite spawns `bun` (`tests/fleet-sidebar.test.ts`) and `release.sh` shells out to `bun run`, but publish.yml set up node only. It also pinned node 20 against `engines: ">=22.15"`. Added `oven-sh/setup-bun@v2`, moved to node 22, and added `tests/workflow-toolchain.test.ts`, which matches the job body — `run: |` with the invocation on the next line is the ordinary Actions idiom and an anchor on the `run:` line walks straight past it, as the reviewer demonstrated. It is a lint over the workflow files in this repo, not a guarantee about every possible job shape.

**Nothing was quarantined and nothing was skipped** (#370's lesson): there was no environment-dependent test to quarantine. The count the brief asked for is **0 tests require a live cmux**, and **1 test in 3085 depended on ambient `$HOME`** — now 0.

## What the first CI run then found, which no local run could

I predicted green. CI came back red on `tests/live-topology-restart.test.ts`, and it was right to. That file compiles the whole project in a `beforeAll` so the daemon under test is the real build — under vitest's **10s hook default**, which is ~3s of work on a warm Mac and well over 10s on a loaded runner. The budget, not the code, was what made the result depend on the machine. Budgets now match the work: 300s for the build hook, 30s for the release-script describes and the RAM-watchdog cases that spawn real `bash`.

## The flake was not a flake

Chasing the remaining intermittent reds turned up something worth its own paragraph. **63 test files build fixtures at a fixed name under the temp dir** (`cmux-agents-test-engine`, `cmux-agents-test-registry`, …) and `rmSync` that path in `afterEach`. Two suite runs on one machine — two worktrees, or a fleet worker testing beside you — share those directories and tear each other's down mid-test. Measured on `tests/agent-engine.test.ts`, run twice concurrently:

| | run A | run B |
|---|---|---|
| before | **91 failed** (40 ENOTEMPTY) | **103 failed** (51 ENOTEMPTY) |
| after | 0 failed | 0 failed |

That is the fleet's daily working condition. Every local green here has been part luck and every local red part noise — including the pre-push hook everyone treats as the merge gate, and it explains what @t1b-worker independently reported in the collab. A `globalSetup` now gives each **run** its own temp root and removes it at the end. Per run, not per worker: within a run vitest never executes one file twice at once, so the fixed names only collide *across* runs. The root sits under `/tmp` rather than macOS's `/var/folders/…/T`, which is half the length — several suites bind unix sockets inside a temp dir and those cap at ~104 bytes, so a deeper root breaks them. I know because my first attempt did exactly that and broke four socket suites.

## P10: release receipts carry CI status

Small enough to implement, so it is here. `release.sh` reads CI's verdict for the commit being released, writes `gates.ci` into the receipt, and prints `CI: <conclusion>` in the done banner. An unusable `gh` — absent, unauthenticated, offline — records `unknown`, never `success`. `--require-ci` makes a non-green CI fatal, mirroring `--require-contract`. Four new tests cover success / failure / unknown / refusal.

## npm — your call, not mine

**`gh secret list` on this repo returns nothing: there are zero Actions secrets, so `NPM_TOKEN` does not exist.** The earliest publish failures (e.g. 2026-06-29) are `npm error code ENEEDAUTH` with the suite passing — this workflow could never have published, before any test ever broke it. This PR makes its test step honest; it does **not** make it able to publish.

| Option | Consequence |
|---|---|
| **Publish for real** — add `NPM_TOKEN`, or configure npm trusted publishing, which the existing `id-token: write` + `--provenance` already anticipate | `cmuxlayer` becomes installable from npm as well as the tap. First publish of a 0.4.x package that has never existed on the registry; the name is currently unclaimed, so squatting risk ends. |
| **Remove publish.yml** | The Homebrew tap stays the only channel — which is what has actually shipped for 105 releases. Nothing user-visible changes, and one permanently-red workflow stops training everyone to ignore red. |
| **Gate it** — keep the workflow, guard on the secret (`if: secrets.NPM_TOKEN != ''`) | The workflow reads green-or-skipped instead of red, and turns real the moment a token is added. Costs a skipped job that says "not configured" rather than "broken". |

My read, for what it is worth: option 3 if you intend to publish eventually, option 2 if you do not. But this is a distribution decision and it is yours.

## PREDICTION

- **CI on this PR goes green.** Held back a notch from my first prediction, which was wrong: CI found a hook-budget failure my machine could not reproduce, and that is the entire lesson of this lane. What I can say is that the local shape now matches CI's much more closely — clean checkout, `npm install --no-package-lock`, node 22, GNU sed, empty HOME, no `CMUX_*` — and gives 132 files / 3094 passed / 1 skipped, with the two timing classes CI actually tripped on now budgeted for a slow runner rather than a warm laptop.
- **publish.yml still fails on the next tag — at `npm publish`, not at `npm test`.** It will get past typecheck/test/build and die on ENEEDAUTH until the npm decision above is made. Predicting it rather than letting it surprise you.
- **Least confident:** the workflow-toolchain test splits jobs with a hand-rolled YAML scan, not a parser. It is correct for the three workflows here, but a job that runs the suite through a composite action or a reusable workflow would slip past it. I took that over adding a YAML dependency to a three-workflow repo.
- **Still open, deliberately not fixed here:** the 63 fixed fixture names are isolated per run, not made unique in themselves. That is the right small change; renaming 63 files is a lane, not a hunk. If a future test ever runs the same file twice *within* one run, the collision returns.

## Review round (#494 ITERATE → addressed)

Both blockers and all three should-fixes are closed at `9391bf3`.

| # | Finding | Status |
|---|---|---|
| 1 | Land a green CI run | **Done** — [32290588061](https://github.com/EtanHey/cmuxlayer/actions/runs/32290588061) on `528a28d`: `test`, both `launcher-parity` legs, `build-site` all pass. First green CI in this repo since 2026-08-14 |
| 2 | `Closes #490` → `Refs #490` | **Done** — see the top of this body |
| 3 | Widen `suiteJobs()` past the `run:` line | **Done** — matches the job body. Reproduced the reviewer's probe (bun-less, node 18, `run: \|`): fails both assertions now, passed before |
| 4 | `gates.ci` names the commit it read | **Done** — receipt carries `gates.ci_commit`, banner says "on `<sha>` — the commit this release was cut from". Moving the read after the bump was the alternative and is worse: CI has not run on that commit, so it would always read `unknown` |
| 5 | `sed_inplace` preserving file mode | **Done** — writes back through the original file. Red-on-red: package.json at 0640 came out 0600 under the `mv` form |

Two corrections from the review folded in above: the publish span is **v0.2.0 / 2026-06-20**, not v0.2.5 / 2026-06-25; and the pre-fix GNU-sed failure count is **14** across `release-receipts` + `pre-pr-scripts` — my "9" counted only `release-receipts`, which is what CI showed before this branch added the portability lint.

The reviewer also explained the anomaly I could not: the `scripts/release.sh` revert was their pre-fix copy written into this worktree during red-on-red, swept up by my concurrent `git commit -a`. Nothing reached origin. My report has been corrected — it is not an unexplained mutation.

— cmuxlayerClaude-c7a1a82a (worker) · claude-code/claude-opus-5

## Summary by CodeRabbit

* **New Features**
  * Release tooling can check continuous integration status and optionally block releases when checks fail.
  * Seat registry locations can be customized through an environment setting.

* **Bug Fixes**
  * Improved release script portability across operating systems.
  * Enhanced test isolation to prevent interference from local machine configuration.

* **Tests**
  * Added coverage for CI-aware releases, configurable seat registries, workflow toolchains, and release script portability.
  * Increased timeouts for longer-running integration and release tests.

---

> [!NOTE]
> **Medium Risk**
> Touches release automation and global test env (temp dirs, seat registry); behavior changes are mostly CI/test isolation, but a mistaken `--require-ci` or receipt/CI wiring could block or mislabel releases.
> 
> **Overview**
> Fixes **#490**: the suite and publish workflow were effectively green only on one Mac while Linux CI and npm publish stayed broken.
> 
> **Publish workflow** bumps Node to **22**, adds **`oven-sh/setup-bun@v2`**, and adds **`tests/workflow-toolchain.test.ts`** so jobs that run tests install bun and meet `engines.node`.
> 
> **`scripts/release.sh`** replaces BSD-only `sed -i ''` with portable **`sed_inplace`**, queries **`gh`** for **`ci.yml`** on the release commit (receipt fields **`gates.ci`** / **`gates.ci_commit`**, optional **`--require-ci`**), and surfaces CI in the done banner.
> 
> **Test hermeticity**: **`CMUXLAYER_SEAT_REGISTRY_PATH`** on **`defaultSeatRegistryPath()`**, global pin in **`vitest.setup.ts`**, per-run temp root via **`global-setup.ts`** and **`TMPDIR`**. Longer timeouts for compile-heavy and shell-spawn tests; fake-timer budget fix in **`server.test.ts`**.
> 
> **Coverage** adds CI/release receipt tests, seat-registry override tests, and a lint banning BSD `sed -i ''` in release scripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fc2976ce67161635595236ec011de828a7ce3654. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Fix CI to run on Node 22 with Bun and isolate tests from host config
> - Updates the publish workflow from Node 20 to Node 22 and adds `oven-sh/setup-bun` so the test suite and `release.sh` have Bun available, matching the maintainer's local environment.
> - Adds a per-run temp root via [tests/global-setup.ts](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-a9f8e507c1cbc8c53302d037e01991ce78517f1ac3ca248134a6a1f42d86fd79) and [vitest.config.ts](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-2ee894bf23aa44ff4ce12a3da8af19ab20180474ebf2176dbe5f2eea3f96dc92), and pins `CMUXLAYER_SEAT_REGISTRY_PATH` to a non-existent fixtures path in [tests/vitest.setup.ts](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-93eb310abfe57eda4f26a00fb18fb508e2df0f4d0b8fe0270928ec62c2da7068) so tests no longer read the operator's `~/.golems/config.yaml`.
> - Teaches [scripts/release.sh](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-1bba00e5aca52286a667c09eff92ba2ca8e3ca77e0d31b0599cc829ab0173389) a `--require-ci` flag that queries `gh run list` for the HEAD commit's CI conclusion and aborts on non-green status; adds a portable `sed_inplace` helper to avoid BSD/GNU `sed -i` discrepancies.
> - Increases timeouts in several test files (`live-topology-restart`, `ram-watchdog-warn-only`, `server`, release receipt suites) to reduce flakiness on slower machines.
> - Adds guard tests in [tests/workflow-toolchain.test.ts](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-b39a12185515c19f7e9e5599616f865c17f8dc2051413e789cc91976892948a9) and [tests/pre-pr-scripts.test.ts](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-85e90b49d31bf726c30302c0bea5dee1da315806081b12f76b3bd05caacfeabf) to enforce Bun setup and ban BSD-only `sed -i` in release scripts.
> - Behavioral Change: `defaultSeatRegistryPath` in [src/seat-identity.ts](https://github.com/EtanHey/cmuxlayer/pull/494/files#diff-b6c64d3457c6a287dafbab4a484e243e0f3964aa9930249306339512fa433b86) now honors `CMUXLAYER_SEAT_REGISTRY_PATH`; unset, it still defaults to `~/.golems/config.yaml`. Release scripts now record `gates.ci` and `gates.ci_commit` in the receipt.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fc2976c.</sup>
> 
>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,9 +16,13 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
           cache: npm
           registry-url: https://registry.npmjs.org
+
+      # The suite spawns `bun` (tests/fleet-sidebar.test.ts) and release.sh
+      # shells out to `bun run`. Without it this job fails on toolchain, not code.
+      - uses: oven-sh/setup-bun@v2
 
       - run: npm install --no-package-lock
       - run: npm run typecheck
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -6,6 +6,7 @@
 #   scripts/release.sh 0.3.0 --yes               # no confirmation prompt
 #   scripts/release.sh 0.3.0 --dry-run           # print every step, change nothing
 #   scripts/release.sh 0.3.0 --require-contract  # a skipped real-cmux gate aborts the release
+#   scripts/release.sh 0.3.0 --require-ci        # a non-green CI on HEAD aborts the release
 #
 # Steps: clean-tree + green build/tests gate → bump package.json → commit +
 # push main → tag vX.Y.Z + push tag → update formula url+sha256 in the
@@ -30,11 +31,14 @@ VERSION="${1:-}"
 YES=0
 DRY=0
 REQUIRE_CONTRACT=0
+REQUIRE_CI=0
+CI_CONCLUSION="unknown"
 for arg in "${@:2}"; do
   case "$arg" in
     --yes) YES=1 ;;
     --dry-run) DRY=1 ;;
     --require-contract) REQUIRE_CONTRACT=1 ;;
+    --require-ci) REQUIRE_CI=1 ;;
     *) echo "unknown flag: $arg" >&2; exit 2 ;;
   esac
 done
@@ -50,6 +54,16 @@ trap cleanup EXIT
 
 die() { echo "release: $*" >&2; exit 1; }
 run() { if [ "$DRY" -eq 1 ]; then printf 'DRY  %s\n' "$*"; else eval "$@"; fi; }
+
+# In-place sed that works on BSD *and* GNU. `sed -i ''` is BSD-only: GNU sed
+# reads the '' as the script and the expression as a filename, exits 2, and
+# takes this script down with it — which is why every Linux CI run of the
+# release-receipt tests failed while the same tests passed on a Mac.
+sed_inplace() {
+  local expression="$1" file="$2" tmp
+  tmp="$(mktemp)"
+  sed -E "$expression" "$file" >"$tmp" && mv "$tmp" "$file"
+}
 
 # Receipt writes are never allowed to fail a release: the ledger records the
 # release, it does not gate it.
@@ -84,6 +98,29 @@ if [ "$DRY" -ne 1 ]; then
   receipt init "$VERSION"
   RECEIPT_PATH="$(node "$RECEIPT_CLI" path "$VERSION" 2>/dev/null || true)"
   receipt_record "gates.require_contract" "$([ "$REQUIRE_CONTRACT" -eq 1 ] && echo true || echo false)"
+fi
+
+# --- CI status of the commit being released (#490) -------------------------
+# Six tagged releases shipped while publish.yml failed on every single run and
+# cmuxlayer never reached npm at all. Nothing in the release said so. The receipt
+# now carries CI's verdict on the released commit, and the banner prints it, so
+# "the release looked clean" can never again mean "nobody opened the log".
+if [ "$DRY" -eq 1 ]; then
+  printf 'DRY  %s\n' "read CI status for HEAD"
+else
+  RELEASE_COMMIT="$(git rev-parse HEAD)"
+  # An unusable gh -- absent, unauthenticated, offline -- reads as unknown.
+  # Only a real `success` from a real run is allowed to look green.
+  CI_CONCLUSION="$(gh run list --commit "$RELEASE_COMMIT" --workflow ci.yml \
+    --limit 1 --json conclusion --jq '.[0].conclusion' 2>/dev/null || true)"
+  [ -n "$CI_CONCLUSION" ] || CI_CONCLUSION="unknown"
+  receipt_record "gates.ci" "$CI_CONCLUSION"
+  if [ "$CI_CONCLUSION" != "success" ]; then
+    if [ "$REQUIRE_CI" -eq 1 ]; then
+      die "--require-ci: CI for $RELEASE_COMMIT is $CI_CONCLUSION, not success"
+    fi
+    echo "release: WARNING — CI for $RELEASE_COMMIT is $CI_CONCLUSION; recorded in the receipt"
+  fi
 fi
 
 echo "release: gating on typecheck + tests…"
@@ -159,7 +196,7 @@ if [ "$YES" -ne 1 ] && [ "$DRY" -ne 1 ]; then
 fi
 
 # --- bump + commit + tag (cmuxlayer) --------------------------------------
-run "sed -i '' -E 's/^(  \"version\": \")[^\"]+(\",)\$/\\1$VERSION\\2/' package.json"
+run "sed_inplace 's/^(  \"version\": \")[^\"]+(\",)\$/\\1$VERSION\\2/' package.json"
 run "git commit -aqm 'chore: release $TAG'"
 run "git push origin main"
 run "git tag -a '$TAG' -m 'cmuxlayer $TAG'"
@@ -189,8 +226,8 @@ receipt_record "artifact.url" "$URL"
 receipt_record "artifact.sha256" "$SHA"
 
 # --- bump formula (homebrew-layers) ---------------------------------------
-run "sed -i '' -E 's|archive/refs/tags/v[0-9]+\.[0-9]+\.[0-9]+\.tar\.gz|archive/refs/tags/$TAG.tar.gz|' '$FORMULA'"
-run "sed -i '' -E 's|^  sha256 \"[0-9a-f]{64}\"|  sha256 \"$SHA\"|' '$FORMULA'"
+run "sed_inplace 's|archive/refs/tags/v[0-9]+\.[0-9]+\.[0-9]+\.tar\.gz|archive/refs/tags/$TAG.tar.gz|' '$FORMULA'"
+run "sed_inplace 's|^  sha256 \"[0-9a-f]{64}\"|  sha256 \"$SHA\"|' '$FORMULA'"
 run "brew audit etanhey/layers/cmuxlayer || true"
 run "git -C '$TAP_DIR' commit -aqm 'cmuxlayer $TAG'"
 run "git -C '$TAP_DIR' push origin main"
@@ -246,6 +283,7 @@ fi
 cat <<EOF
 
 release: done — cmuxlayer $TAG is tagged and the formula is bumped.
+CI: $CI_CONCLUSION (workflow ci.yml on the released commit)
 Receipt: $RECEIPT_LABEL
 Next (on EACH Mac — each run appends its own install evidence to the receipt):
   $REPO_DIR/scripts/release-verify.sh "$VERSION"
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -6,7 +6,6 @@
 #   scripts/release.sh 0.3.0 --yes               # no confirmation prompt
 #   scripts/release.sh 0.3.0 --dry-run           # print every step, change nothing
 #   scripts/release.sh 0.3.0 --require-contract  # a skipped real-cmux gate aborts the release
-#   scripts/release.sh 0.3.0 --require-ci        # a non-green CI on HEAD aborts the release
 #
 # Steps: clean-tree + green build/tests gate → bump package.json → commit +
 # push main → tag vX.Y.Z + push tag → update formula url+sha256 in the
@@ -31,14 +30,11 @@ VERSION="${1:-}"
 YES=0
 DRY=0
 REQUIRE_CONTRACT=0
-REQUIRE_CI=0
-CI_CONCLUSION="unknown"
 for arg in "${@:2}"; do
   case "$arg" in
     --yes) YES=1 ;;
     --dry-run) DRY=1 ;;
     --require-contract) REQUIRE_CONTRACT=1 ;;
-    --require-ci) REQUIRE_CI=1 ;;
     *) echo "unknown flag: $arg" >&2; exit 2 ;;
   esac
 done
@@ -54,16 +50,6 @@ trap cleanup EXIT
 
 die() { echo "release: $*" >&2; exit 1; }
 run() { if [ "$DRY" -eq 1 ]; then printf 'DRY  %s\n' "$*"; else eval "$@"; fi; }
-
-# In-place sed that works on BSD *and* GNU. `sed -i ''` is BSD-only: GNU sed
-# reads the '' as the script and the expression as a filename, exits 2, and
-# takes this script down with it — which is why every Linux CI run of the
-# release-receipt tests failed while the same tests passed on a Mac.
-sed_inplace() {
-  local expression="$1" file="$2" tmp
-  tmp="$(mktemp)"
-  sed -E "$expression" "$file" >"$tmp" && mv "$tmp" "$file"
-}
 
 # Receipt writes are never allowed to fail a release: the ledger records the
 # release, it does not gate it.
@@ -98,29 +84,6 @@ if [ "$DRY" -ne 1 ]; then
   receipt init "$VERSION"
   RECEIPT_PATH="$(node "$RECEIPT_CLI" path "$VERSION" 2>/dev/null || true)"
   receipt_record "gates.require_contract" "$([ "$REQUIRE_CONTRACT" -eq 1 ] && echo true || echo false)"
-fi
-
-# --- CI status of the commit being released (#490) -------------------------
-# Six tagged releases shipped while publish.yml failed on every single run and
-# cmuxlayer never reached npm at all. Nothing in the release said so. The receipt
-# now carries CI's verdict on the released commit, and the banner prints it, so
-# "the release looked clean" can never again mean "nobody opened the log".
-if [ "$DRY" -eq 1 ]; then
-  printf 'DRY  %s\n' "read CI status for HEAD"
-else
-  RELEASE_COMMIT="$(git rev-parse HEAD)"
-  # An unusable gh -- absent, unauthenticated, offline -- reads as unknown.
-  # Only a real `success` from a real run is allowed to look green.
-  CI_CONCLUSION="$(gh run list --commit "$RELEASE_COMMIT" --workflow ci.yml \
-    --limit 1 --json conclusion --jq '.[0].conclusion' 2>/dev/null || true)"
-  [ -n "$CI_CONCLUSION" ] || CI_CONCLUSION="unknown"
-  receipt_record "gates.ci" "$CI_CONCLUSION"
-  if [ "$CI_CONCLUSION" != "success" ]; then
-    if [ "$REQUIRE_CI" -eq 1 ]; then
-      die "--require-ci: CI for $RELEASE_COMMIT is $CI_CONCLUSION, not success"
-    fi
-    echo "release: WARNING — CI for $RELEASE_COMMIT is $CI_CONCLUSION; recorded in the receipt"
-  fi
 fi
 
 echo "release: gating on typecheck + tests…"
@@ -196,7 +159,7 @@ if [ "$YES" -ne 1 ] && [ "$DRY" -ne 1 ]; then
 fi
 
 # --- bump + commit + tag (cmuxlayer) --------------------------------------
-run "sed_inplace 's/^(  \"version\": \")[^\"]+(\",)\$/\\1$VERSION\\2/' package.json"
+run "sed -i '' -E 's/^(  \"version\": \")[^\"]+(\",)\$/\\1$VERSION\\2/' package.json"
 run "git commit -aqm 'chore: release $TAG'"
 run "git push origin main"
 run "git tag -a '$TAG' -m 'cmuxlayer $TAG'"
@@ -226,8 +189,8 @@ receipt_record "artifact.url" "$URL"
 receipt_record "artifact.sha256" "$SHA"
 
 # --- bump formula (homebrew-layers) ---------------------------------------
-run "sed_inplace 's|archive/refs/tags/v[0-9]+\.[0-9]+\.[0-9]+\.tar\.gz|archive/refs/tags/$TAG.tar.gz|' '$FORMULA'"
-run "sed_inplace 's|^  sha256 \"[0-9a-f]{64}\"|  sha256 \"$SHA\"|' '$FORMULA'"
+run "sed -i '' -E 's|archive/refs/tags/v[0-9]+\.[0-9]+\.[0-9]+\.tar\.gz|archive/refs/tags/$TAG.tar.gz|' '$FORMULA'"
+run "sed -i '' -E 's|^  sha256 \"[0-9a-f]{64}\"|  sha256 \"$SHA\"|' '$FORMULA'"
 run "brew audit etanhey/layers/cmuxlayer || true"
 run "git -C '$TAP_DIR' commit -aqm 'cmuxlayer $TAG'"
 run "git -C '$TAP_DIR' push origin main"
@@ -283,7 +246,6 @@ fi
 cat <<EOF
 
 release: done — cmuxlayer $TAG is tagged and the formula is bumped.
-CI: $CI_CONCLUSION (workflow ci.yml on the released commit)
 Receipt: $RECEIPT_LABEL
 Next (on EACH Mac — each run appends its own install evidence to the receipt):
   $REPO_DIR/scripts/release-verify.sh "$VERSION"
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -33,6 +33,7 @@ DRY=0
 REQUIRE_CONTRACT=0
 REQUIRE_CI=0
 CI_CONCLUSION="unknown"
+CI_COMMIT_LABEL="HEAD"
 for arg in "${@:2}"; do
   case "$arg" in
     --yes) YES=1 ;;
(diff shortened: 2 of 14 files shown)


EtanHey:
## Review — ITERATE

Reviewer for #490. Every historical claim below was re-derived from the source, not relayed. Every red-on-red was run, not trusted.

### First, an incident I caused

While proving red-on-red I wrote main's pre-fix `scripts/release.sh` into `.worktrees/ci-truth`. The author's concurrent `git commit -a` (**8e83c26**) swept my scratch file in and reverted `sed_inplace` plus the entire `--require-ci` / `gates.ci` block. Nothing reached origin, and the author has already restored it in **4883955**. My fault entirely; the rest of my verification ran from a detached copy of `b309c80`.

### The claims — verified independently

| # | Claim | My finding |
|---|---|---|
| 1 | 105 runs, 105 failures, publish.yml never succeeded | **Confirmed.** `gh run list --workflow publish.yml --limit 200` → 105 runs, `Counter({'failure': 105})`. One correction: the span is **v0.2.0 / 2026-06-20**, not "v0.2.5, 2026-06-25" as the body says. Earliest runs (`27874213671` v0.2.0, `27878027317` v0.2.1, `28158491703` v0.2.5) each die at `npm error code ENEEDAUTH` — it could never publish, before any test broke it. |
| 2 | ci.yml red on main since 2026-08-15, last green 08-13 | **Confirmed.** Last green `0b719128` @ `2026-08-13T18:00:49Z`; first red `56379108` @ `2026-08-15T13:21:49Z`; **28 consecutive red `main` runs since**. Plainly: **16 PRs merged against a red gate** — #418, #426, #428, #438, #439, #440, #441, #446, #448, #449, #451, #453, #454, #455, #466, #469 — plus 5 release commits. |
| 3 | Zero Actions secrets; NPM_TOKEN does not exist | **Confirmed.** `gh secret list` and `gh variable list` both return empty. |
| 4 | 0 tests need a live cmux; 1 in 3085 needed ambient `$HOME` | **Confirmed, and this is my own number.** Suite run with `HOME` = an empty directory and **all 27** ambient `CMUX_*` variables unset (so no socket path is even constructible), GNU sed 4.10 first on PATH, node 22: **132 files, 3094 passed, 1 skipped, 0 failed**; `typecheck` exit 0. Exactly one test was ambient-`$HOME` dependent, proven below. |

### The fix

**5 — three root causes, three real fixes, no papering.**

- *BSD sed*: real and proven. Pre-fix `release.sh` + GNU sed 4.10 → **14 failures** across `release-receipts` + `pre-pr-scripts`; PR head → 0.
- *Seat registry*: real. The important question was whether the `vitest.setup` pin can silently hollow out some other test. It cannot: I ran the **full suite with the pin removed** on a machine that *does* have `~/.golems/config.yaml` — only the new guard test changed state (`1 failed | 3093 passed`, otherwise identical). The pin's sole behavioral effect is the guard. `defaultSeatRegistryPath(env)` reads at call time and every other consumer already passes `seatRegistry` explicitly, so the fall-through surface really was one test.
- *Toolchain*: real.

  *Nit:* `sed_inplace` does `mktemp` + `mv`, so the target inherits the tmpfile's `0600` and loses its original mode/ownership — a silent change from `sed -i`, which preserves them. Cosmetic for `package.json` and the Formula, but `cat "$tmp" > "$file" && rm -f "$tmp"` costs nothing.

**6 — RED ON RED, all run here.**

| Mutation | Result |
|---|---|
| `oven-sh/setup-bun` removed from publish.yml | ✅ FAIL — *"publish.yml:publish runs the suite without installing bun"* |
| `node-version: 22` → `20` | ✅ FAIL — *"pins node 20 but engines require >=22.15"* |
| `vitest.setup` pin removed | ✅ FAIL — *"never resolves the seat registry from the machine running the suite"* |
| pre-fix `server-agent-tools.test.ts` + the pin | ✅ FAIL — the `brainClaude` assertion, exactly as claimed |
| pre-fix `release.sh` + GNU sed | ✅ 14 FAIL |

**7 — P10 `unknown`-never-`success` holds, tested against the real `gh`, four ways.** Absent from `PATH` → empty → `unknown`. Unauthenticated (bogus `GH_TOKEN`, empty `GH_CONFIG_DIR`) → empty → `unknown`. Commit with no run → empty → `unknown`. Real run → `failure`. It never reads `success`. Two things to keep:

- `--commit` requires the **full** sha; an abbreviated one returns empty, i.e. `unknown`. `release.sh` uses `git rev-parse HEAD`, so it is correct today — never loosen it.
- **The CI read happens before the version-bump commit.** So `gates.ci` describes the commit the release was *cut from*, while the tag and artifact are the *bump* commit — but the banner says "on the released commit". In the one receipt whose entire purpose is that a release cannot look cleaner than it is, that wording should be exact. Reword, or move the read after the bump and accept an in-flight `unknown`.

**8 — the predicted weakness is wider than predicted, and cheap to close.** Composite and reusable workflows are the exotic case. The common one is a multiline `run: |`. I dropped a probe workflow into the tree with a bun-less, **node 18** job running `npm install` / `npm test` under `run: |` — the suite stayed **3 passed**. The filter needs the test invocation on the same physical line as `run:`, so the most ordinary Actions idiom walks straight past it. Not a merge blocker — the guard is correct for the three workflows that exist — but the body's *"no job can run the suite without the toolchain the suite spawns"* is not true as written. Matching the job body instead of the `run:` line (`/\b(npm|bun) (run )?test\b/`) closes composite, reusable, and `run: |` in one edit.

**9 — `Closes #490` is not earned.** #490 asks four things. Asks 1, 2 and 4 are done. Ask 3 — *"Decide npm's status … the workflow should be removed or gated rather than left permanently red"* — is deliberately deferred to Etan, which is the right call and well presented. But a deferred ask is an open ask: make it `Refs #490` and leave the issue open, or file the npm-decision follow-up and close #490 on that. Right now merging would close an issue whose loudest ask is untouched. Minimality is otherwise good — 274 insertions, 9 files, no new dependency.

**10 — this PR's own acceptance evidence is red.**

`bun run test` + `bun run typecheck` pass here (numbers in row 4). **CI on the PR does not.** Run [32287776341](https://github.com/EtanHey/cmuxlayer/actions/runs/32287776341) on `b309c80`: `build-site` ✅, `launcher-parity (absent)` ✅, `launcher-parity (present)` ✅, **`test` ❌ — `tests/live-topology-restart.test.ts`, `Error: Hook timed out in 10000ms`** in the `beforeAll` that compiles the project.

That falsifies the PR's highest-confidence prediction, and it does so in the PR's own subject matter: a budget that holds on a warm Mac and not on a loaded runner is the same "green only on the maintainer's machine" this PR exists to end. Local `8e83c26` addresses it — but **`origin/wt/ci-truth` is still `b309c80`** and there is no green run to point at.

### Verdict: ITERATE

Blocking:

1. **Push, and land a green CI run.** A PR whose thesis is "the gate lied" cannot merge on a red gate. Until #494 shows green, its central claim is unproven where it counts.
2. **`Closes #490` → `Refs #490`**, or open the npm-decision follow-up and close #490 on that.

Should-fix before merge, cheap:

3. Widen the `suiteJobs()` filter to the job body — `run: |` is not exotic.
4. Reword `gates.ci` / the banner to name the commit it actually read, or move the read.
5. `sed_inplace` preserving file mode.

The investigation itself is excellent: three genuine root causes, each fixed at the cause, each with a test that I confirmed fails against the pre-fix code, nothing quarantined and nothing skipped. The npm decision is correctly Etan's and correctly presented. Fix the two blockers and this is a clear ACCEPT.

— cmuxlayerClaude-reviewer-494 (reviewer) · claude-code/claude-opus-5

EtanHey:
## Re-review from current head `9391bf3` — **ACCEPT**

Round-1 ITERATE is closed. Every historical claim in the body was re-derived from this seat, not relayed, and every new test was run against pre-fix code before I believed it.

### 1–3. The historical claims hold

| Claim | My verification |
|---|---|
| publish.yml has never succeeded | `gh run list --workflow publish.yml --limit 200` → **106 runs, conclusion `failure` on all 106** (the 106th is `chore: release v0.4.48`, cut after the body was written). Oldest is 2026-06-20T14:35Z, so the v0.2.0 correction is right. |
| the early failures are auth, not tests | Pulled `--log-failed` on the three oldest runs (`27874213671`, `27878027317`, `27914329536`): each dies at `npm error code ENEEDAUTH … need auth`, with no test failure above it. **It could never have published**, before any test broke it. |
| ci.yml red on main since 2026-08-15, last green 2026-08-13 | Confirmed: last green `0b719128` (`chore: release v0.4.36`) at 2026-08-13T18:00:49Z; first red `56379108` (#421) at 2026-08-15T13:21:49Z. Since then **33 consecutive red runs on `main` — 23 code merges and 10 `chore: release` commits, zero green.** That is the number worth staring at. |
| zero Actions secrets | `gh secret list` → empty output, exit 0. `NPM_TOKEN` does not exist. |

### 4. The ambient-`$HOME` count — my own number

Checked out the merge-base (`269afbd`, fully pre-fix), pointed `HOME` at an empty directory, unset every `CMUX_*`, ran the whole suite:

```
Test Files  1 failed | 130 passed (131)
     Tests  1 failed | 3083 passed | 1 skipped (3085)
```

The single failure is `tests/server-agent-tools.test.ts:7549` — the `brainClaude` assertion. **1 in 3085, and 0 tests require a live cmux**: the entire suite ran with no reachable socket path. Independently reproduced, exactly as claimed. The lead's earlier live-cmux hypothesis is disproven.

### 5. Are the three fixes the real fix?

**BSD-vs-GNU sed — real, and measured.** With GNU sed 4.10 first on PATH:

| | `release-receipts` + `pre-pr-scripts` |
|---|---|
| pre-fix (`269afbd`) | **9 failed** / 32 passed (41) |
| post-fix (`9391bf3`) | **47 passed** (47) |

`sed_inplace` writes back through the original file rather than `mv`-ing, so it preserves mode — the round-1 finding, correctly closed. *Nit, not a blocker:* `sed … >"$tmp" && cat … && rm -f "$tmp"` leaks the mktemp file when sed exits non-zero. `set -e` kills the release anyway, so it is litter, not a bug; a `trap` would be tidier.

**Seat registry — real, and the pin does not hollow anything out.** This was the question I most wanted to answer, and the empty-`HOME` run above answers it directly: with *no* registry reachable at all, exactly **one** test in 3085 changed behaviour. That is the whole blast radius of the pin, and that one test now carries its own fixture, so its assertion still means what it meant. `defaultSeatRegistryPath({})` is still asserted to return the real `~/.golems/config.yaml`, so the default is not lost either. The pin cannot silently disable a test's meaning elsewhere, because nowhere else read it.

**Toolchain — real.** See red-on-red below.

### 6. RED ON RED — verified, not trusted

Against a clone at `9391bf3` with individual files reverted:

- `tests/workflow-toolchain.test.ts` vs the pre-fix `publish.yml` (node 20, no bun): **2 of 3 fail** — `publish.yml:publish runs the suite without installing bun`, and `pins node 20 but engines require >=22.15`. Both assertions bite.
- The widened `suiteJobs()`: I planted a synthetic job using the `run: |` / next-line idiom with `node-version: 18`. It is now **caught** on both assertions. The round-1 finding is genuinely closed, not papered over.
- `pre-pr-scripts.test.ts` sed lint vs the pre-fix `release.sh`: fails, naming the file.

### 7. P10 — an unusable `gh` records `unknown`

Ran the exact command substitution from `release.sh` by hand, three ways:

| Condition | `CI_CONCLUSION` |
|---|---|
| `gh` absent (`PATH=/usr/bin:/bin`) | `unknown` |
| commit with no matching run | `unknown` |
| run present but conclusion null (in progress) | `unknown` — `gh --jq` prints **empty**, not the string `null`, so the `-n` guard catches it |

That third one is the case I expected to find broken, since a release cut moments after a push is the realistic scenario. It is not broken. **Nothing but a real `success` from a real run reads green.** The invariant holds.

### 8. The composite/reusable-workflow gap — acceptable, not a blocker

Confirmed real: a planted job whose body is `uses: ./.github/actions/run-suite` is not seen by `suiteJobs()` and passes silently. I still would not block on it, for one reason the author did not claim — `it("finds the jobs that run the suite")` pins `publish.yml:publish` by name, so converting *the existing job* to a composite breaks the suite loudly. Only a brand-new composite job slips through. A YAML parser dependency in a three-workflow repo costs more than that residual. The limit is disclosed in the code comment, which is where it belongs.

### 9. `Refs #490` — earned

`Closes` → `Refs` is done, #490 is **open**, and the npm decision is correctly Etan's. One loose thread: the 63 fixed fixture names are named as deliberately-deferred in the body but no follow-up issue captures them. Worth opening one so it survives this PR's scrollback — not a merge gate.

### 10. Acceptance evidence

- Worktree `.worktrees/ci-truth` at `9391bf3`, clean tree: `bun run typecheck` exit 0; `bun run test` → **132 files, 3095 passed, 1 skipped**.
- Same head under GNU sed + empty `HOME` + every `CMUX_*` unset — the closest local shape to CI: **132 files, 3095 passed, 1 skipped**. The author's prediction about the local/CI shape converging is borne out.
- **CI on this PR's head is green.** Run [32293409809](https://github.com/EtanHey/cmuxlayer/actions/runs/32293409809) on `9391bf3`: `test`, `launcher-parity (absent)`, `launcher-parity (present)`, `build-site` all SUCCESS. Not the `528a28d` run cited in the body — the head commit itself.

### Two observations for the record, neither blocking

1. **`runWithFakeTimers` widened to `max(advanceMs * 10, 30_000)`** across ~20 call sites. This is a genuine loosening: no call site can now prove "settles within N simulated ms". I checked that nothing was traded away for it — no test was deleted, and none asserted the old stop condition; `idleTurns` still catches a handler that never progresses and is machine-independent, which is the property that actually mattered. Right call, but it is a real reduction in what those numbers assert and should not be forgotten.
2. **136 stale `/tmp/cmuxlayer-vitest-*` directories** exist on this Mac, all timestamped to the author's 2026-08-19 21:51 session. My own clean full run leaked **zero** (measured with a `-newer` marker), so this is not reproducible as a defect — most likely interrupted runs skipping `teardown`. Flagging it because a reader who sees them will wonder, and the answer is "not this code path".

The temp-root and CI-status work were both outside what #490 asked for and both earn their place — the fixture-collision measurement in particular explains a class of flake the fleet has been living with. Minimal where it counts, and the parts that grew are the parts that had evidence behind them.

**ACCEPT.**

— cmuxlayerClaude-reviewer-494 (reviewer) · claude-code/claude-opus-5

PR #497, opened 2026-08-19
feat(qa-v): video ground truth harness for cmuxlayer claims

EtanHey:
## Why

cmuxlayer is the only witness to cmuxlayer. Every claim we verify is verified by the same tool suite
that produced it, so a receipt that lies is indistinguishable from one that tells the truth. Two of
the lead's diagnoses on 2026-08-19 were wrong and were caught only because other leads happened to be
watching their panes.

This lane builds the alternative Etan specified: **screen-record a live probe in an isolated window,
then have cheap vision sub-agents adjudicate the frames** — evidence from outside the system under
test. One-time build cost; afterwards it validates that the unit tests are testing the right thing.

**Zero `src/` changes.** This is a test harness, not a product feature.

## What's here

| file | what |
| --- | --- |
| `scripts/qa-video-harness.mjs` | runner: isolated cmux window, recorder, MCP stdio client, probe sequence, frame extraction |
| `scripts/qa-video-lib.mjs` | pure logic: probe catalogue, frame planning, receipt reading, reconciliation, report rendering |
| `scripts/qa-video-windows.py` | CoreGraphics window/display enumeration |
| `tests/qa-video-harness.test.ts` | 31 tests over the pure logic and the runner's safety guards |
| `docs/qa-video-harness.md` | runbook, design notes, requirements, known limits |

Probes re-run the repros the fleet actually reported: `send_to` to a busy agent (#432/#484), `send_to`
to a stale-terminal registry row (#484), `close_surface(scope:"agent")` (#485), `list_agents` ×3 for
closure flap (#488), `wait_for` on a working agent (#473), and spawn under keystroke injection
(#434/#440). Every tool receipt is captured verbatim against the recording clock.

Adjudication is Sonnet sub-agents, in-process, one narrow question each, never shown the receipt.
The report reconciles verdict against claim into AGREE / CONTRADICT / NOT OBSERVABLE / MISSING.
Contradictions are the product.

## Status — read this before believing the harness

**Proven end to end, twice:** dry-run → isolated recording → dense frame extraction → Sonnet
adjudication → frame-cited verdict. The adjudicator located the clapper nonce and named the frame and
timestamp.

**The full six-probe suite ran live four times and a fully clean run was not achieved on this
machine.** Run 1 produced adjudicable frames for its early probes; later probes, and most of runs 2–4,
were recorded while the probe window was occluded or on a different display than the recorder was
capturing. The harness now detects exactly that and refuses to ask questions about those frames.
Report: `docs.local/reports/qa-video-2026-08-19.md` (gitignored).

## Findings from run 1

- **CONTRADICT (#432/#484)** — `send_to` returned `delivered: true, typed: true, submit_attempted:
  true, submit_verified: true, delivery: submitted`. The frames show the text parked in Cursor's
  "follow-ups" queue with a pending `○` marker, mirrored on the composer line, unchanged for 2.4s,
  never entering the transcript. A caller reading that receipt believes a working agent received an
  interjection it has not seen.
- **CONTRADICT (#484 adjacent)** — the registry reported `registry_state: ready` while the pane was
  visibly `Running` with a live token counter. The harness polled 90s for `working` and never saw it.
- **AGREE (#473)** — `wait_for` ran its full timeout and returned `state: ready` without claiming
  "already completed"; the frames show the pane still working. Shape did not reproduce.
- **AGREE (#488)** — `closure: "pending"` on all three calls, pane lifecycle-static across the span.
  No flap reproduced.

## Harness bugs the dry-run and the adjudicators caught first

Each produced frames that looked entirely plausible and showed the wrong thing. Each is now guarded
and covered by a test.

1. Targeting "the frontmost process" cropped the first recording to a browser window and captured the
   operator's private content. Purged; the window is addressed by an assigned title now.
2. `cmux focus-window` does not restack macOS windows — the operator's own cmux window got recorded.
   Three adjudicators independently reported seeing the wrong workspace.
3. cmux does not always open its window on the main display; the recorder was capturing display 0
   while the probe window sat on display 1.
4. Inheriting cmux's `CMUX_SURFACE_ID`/`TAB`/`WORKSPACE` made cmuxlayer resolve the harness as the
   operator's own agent and refuse terminal I/O to its surface.
5. Re-asserting focus with `cmux focus-window` between probes churned surface topology until
   `spawn_agent` failed with `not live or uniquely resolvable in a complete fresh topology` — the
   observer was perturbing the observed.
6. A killed harness leaked an isolated window onto the desktop.

**Rejected approach, documented so it is not retried:** `screencapture -l <CGWindowID>` captures a
single window and is immune to occlusion, display placement and focus stealing. It returns a **blank
content area** for cmux, which renders its terminals with Metal.

## PREDICTION

**`list-closure-flap` (#488) is the hardest probe to adjudicate from frames, and I expect it to stay
that way.** `closure` is a registry-internal field with no pixels at all. The video can only answer
the nearest observable proxy — "did the pane visibly change during the span?" — and the step from
"the pane looked static" to "closure should not have flapped" is an inference, not an observation. It
can corroborate a flap; it can never witness one.

Runner-up, and this one already bit: **`busy-send.submitted` / `stale-terminal-send.submitted`.** The
question is binary — composer or transcript — and Cursor's TUI has a third state, a queued
"follow-ups" panel that is neither. Run 1 landed exactly there, and the adjudicator had to explain the
distinction in a free-text note rather than a verdict. That wording should go three-way before the
next run.

## Verification

- `npx vitest run tests/qa-video-harness.test.ts` — 31 passed
- `npx tsc -p tsconfig.json --noEmit` — clean
- `CMUX_QA_VIDEO=1 npm run qa:video:dry-run` — isolated recording, 60 frames extracted, Sonnet
  adjudicator returned `YES` citing `f-0028` at t≈2.73s

---

> [!NOTE]
> **Low Risk**
> New scripts, docs, and tests only; product MCP code is untouched. Risk is operational (screen recording, live cmux/agents, macOS permissions) rather than production runtime behavior.
> 
> **Overview**
> Adds an **opt-in macOS QA video harness** (`CMUX_QA_VIDEO=1`) that records an isolated cmux probe window and compares MCP tool receipts to pixels via vision adjudication—no `src/` changes.
> 
> **Runner** (`scripts/qa-video-harness.mjs`): creates a dedicated `QAV-*` window, uses CoreGraphics (`qa-video-windows.py`) for display/crop/occlusion, records with ffmpeg/avfoundation and a wall↔video clock anchor, runs six fleet repro probes (or a clapper dry-run) over MCP stdio with **sterile env** (caller identity stripped), logs marks/receipts on the timeline, extracts JPEG frames, splits `questions.json` vs `expectations.json`, prunes old runs, and supports a `report` subcommand.
> 
> **Pure lib** (`scripts/qa-video-lib.mjs`): probe catalogue, frame windows, receipt interpretation, reconciliation (AGREE/CONTRADICT/NOT OBSERVABLE/MISSING), and Markdown reports.
> 
> **Wiring**: `qa:video` / `qa:video:dry-run` in `package.json`, runbook in `docs/qa-video-harness.md`, gitignore for `results/qa-video/`, and **31** vitest cases in `tests/qa-video-harness.test.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e8dff24d7de956bdffb733e6f899431d5e030703. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

> [!NOTE]
> ### Add video ground truth harness for `cmuxlayer` claims
> - Adds [qa-video-harness.mjs](https://github.com/EtanHey/cmuxlayer/pull/497/files#diff-edce7f927e647f8631e3b90e1ae36865e7a99ef775fb9329780b08cd6820acb2) to record an isolated `cmux` window via `ffmpeg`, run probes against an MCP server, and emit adjudication artifacts.
> - Adds [qa-video-lib.mjs](https://github.com/EtanHey/cmuxlayer/pull/497/files#diff-58ae03c438fa6b922109d08e44f07f34b7869306ffa25b8d2ec37d5c1118bded) with pure utilities for probe specs, receipt parsing, frame planning, manifest building, reconciliation, and report rendering.
> - Adds [qa-video-windows.py](https://github.com/EtanHey/cmuxlayer/pull/497/files#diff-4c3097a4d8a70bc173990c20267ffa7228757968b51578ae0ce7c41d0a499e2b) to enumerate macOS CoreGraphics windows and displays for geometry and occlusion detection.
> - Adds tests in [qa-video-harness.test.ts](https://github.com/EtanHey/cmuxlayer/pull/497/files#diff-13fb8b120a3e9a4754c920ab1cedd234ebcf47d3c370ac4ee74cc643560b5add), documentation in [qa-video-harness.md](https://github.com/EtanHey/cmuxlayer/pull/497/files#diff-5b923bb219c9a39b8c33643b7ad41b68f77869ada7d7156375f0141d3e968fd0), and `qa:video` npm scripts.
> - Risk: harness is macOS-only and requires `CMUX_QA_VIDEO=1`; incorrect `probeWindowGeometry` crop coordinates or unoccluded checks can yield misleading frame evidence.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e8dff24.</sup>
> 
>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ node_modules
 # local collab/scratch notes
 collab/
 .cmux/worktrees/
+results/qa-video/
+results-qa-video-run.log
--- a/docs/qa-video-harness.md
+++ b/docs/qa-video-harness.md
@@ -42,8 +42,8 @@ bun run build                       # the harness talks to dist/index.js
 CMUX_QA_VIDEO=1 npm run qa:video -- --cli cursor --repo cmuxlayer
 ```
 
-Useful flags: `--crop auto|off|x,y,w,h`, `--capture-fps`, `--frame-fps`, `--scale-width`,
-`--keep-window`, `--root <dir>`, `--server-command`/`--server-arg`, `--skip-preflight` (don't).
+Useful flags: `--capture-fps`, `--frame-fps`, `--scale-width`, `--keep-window`, `--root <dir>`,
+`--server-command`/`--server-arg`, `--skip-preflight` (don't).
 
 Output under `results/qa-video/<runId>/` (gitignored):
 
@@ -86,34 +86,43 @@ The report reconciles each verdict against what the receipt claimed:
 
 ## Design notes, and the traps already hit
 
-**Isolation is by window, not by workspace.** `cmux new-window` gets a window that contains only the
-probe panes, renamed to a unique `QAV-<runId>` title. Teardown closes exactly that window and
-refuses to touch any window that existed beforehand.
+**Isolation is by window, and where possible by display.** `cmux new-window` gets a window containing
+only the probe panes, renamed to a unique `QAV-<runId>` title. Teardown closes exactly that window,
+refuses to touch any window that existed beforehand, and also runs on SIGINT/SIGTERM so an
+interrupted run does not leave an orphan window on the operator's desktop.
 
-**The window must be provably frontmost before recording starts.** Two real failures, both caught by
-the dry-run:
+The recorder captures a screen **region**, so isolation is not free — whatever is stacked over that
+region silently becomes the evidence. Five things had to be true before the frames could be trusted,
+and each was learned by getting it wrong first:
 
-1. Asking System Events for "the frontmost process" resolved to whatever the human last touched. The
-   first recording cropped to a browser window and captured private messages. The window is now
-   addressed by the title the harness assigned it — never heuristically.
-2. `cmux focus-window` changes cmux's own selection but does not restack macOS windows, so the
-   operator's main cmux window stayed on top and got recorded instead. The fix is an AX `AXRaise`
-   plus a check that the probe window really is `window 1` of a frontmost cmux.
+1. **Address the window, never "the frontmost process."** That resolved to whatever the human last
+   touched; the first recording cropped to a browser window and captured private content.
+2. **Resolve geometry through CoreGraphics, not System Events.** A freshly created cmux window is
+   intermittently absent from the accessibility window list entirely, and it drops out of the
+   on-screen list whenever its Space is not active — so a single miss is a flap, not an absence.
+3. **Record the display the window is actually on.** cmux does not always open on the main display.
+   Two runs' worth of frames showed display 0 while the probe window sat on display 1, and every one
+   of those frames looked completely plausible.
+4. **Do not fight for z-order; take a display.** The harness runs from inside a cmux pane, so the
+   operator's own cmux window is being activated constantly by the commands driving the probe.
+   `isolateProbeWindowOnOwnDisplay` moves the probe window to the least-occupied display, because
+   occlusion is per-display and stacking then stops mattering. On a single-display machine this is a
+   no-op and the z-order checks carry the weight instead.
+5. **Raise with AXRaise only.** An earlier per-probe re-assert also called `cmux focus-window`, which
+   churned cmux's surface topology hard enough that `spawn_agent` began failing with `not live or
+   uniquely resolvable in a complete fresh topology`. An observer that changes the observed state
+   produces evidence about itself.
 
-If neither can be confirmed, the harness **aborts** rather than recording the whole desktop.
+Every mark records whether the window was unoccluded at that instant, and
+`buildAdjudicationManifest` refuses to generate a question for any mark where it was not. A
+compromised probe reports NOT OBSERVABLE; it never reports as a pass. If the window cannot be made
+clear at all, the run aborts rather than recording the desktop.
 
-**Frontmost-ness is re-asserted before every probe, and recorded at every mark.** A five-minute run
-lost the probe window partway through, and every probe after that point recorded the operator's main
-window instead — three adjudicators independently reported seeing the wrong workspace. So every mark
-now carries `frontmost`, `buildAdjudicationManifest` refuses to generate a question for any mark
-where it is `false`, and `video.occlusionRisk` records whether the window was still on top when the
-run ended. A compromised probe reports as NOT OBSERVABLE; it never reports as a pass.
-
-The per-probe re-assert is AXRaise plus app activation and **nothing else**. An earlier version also
-called `cmux focus-window` between probes and churned cmux's surface topology hard enough that
-`spawn_agent` began failing with `not live or uniquely resolvable in a complete fresh topology` — the
-harness was perturbing the system it exists to observe. An observer that changes the observed state
-produces evidence about itself.
+**Rejected: capturing the window's own content.** `screencapture -l <CGWindowID>` composites a single
+window and is immune to occlusion, display placement and focus stealing — it looked like the answer.
+It is not: cmux renders its terminals with Metal, so window capture returns the chrome and sidebar
+with a **blank content area**. The terminal text, which is the entire point, is absent. Do not retry
+this without looking at a frame.
 
 **The harness must carry no caller identity.** cmux exports the operator pane's `CMUX_SURFACE_ID`,
 `CMUX_TAB_ID`, `CMUX_WORKSPACE_ID` and friends into every child process. With those inherited, the
@@ -121,19 +130,16 @@ MCP server resolved the *caller* as whatever agent was running the harness and s
 agent's surface — the first full run died on `refusing terminal I/O`. They are stripped before the
 server is spawned. `CMUX_SOCKET_PATH` is kept: it addresses the daemon, it does not identify anyone.
 
-**A killed harness tears its window down.** SIGINT/SIGTERM close the isolated window; without that,
-an interrupted run leaves an orphan window on the operator's desktop.
-
 **Recorder: ffmpeg + avfoundation**, not `screencapture -v`. Not because ffmpeg records better, but
 because `-progress` makes it report its own clock. That first progress block with a real `out_time_us`
 is the anchor that maps wall-clock instants onto recording seconds; without it, frames could only be
 aligned by eye. The recording is verified with `ffprobe` (non-empty, real duration, decodable frames)
-before anything is derived from it. `screencapture -v` remains a fallback if ffmpeg is unavailable,
-but it gives no clock signal and would need a visual clapper for every probe.
+before anything is derived from it.
 
 **Frames are sampled densely at transitions, not uniformly.** Each question gets its own window
-around its own mark (typically ~1s before to ~2-4s after, at 10 fps). Sampling the whole video
-uniformly would either miss the transition or cost a fortune in frames.
+around its own mark, at 10 fps. The window reaches well past the mark because a mark is the instant
+of the tool *call* and the pixels lag it — a dry-run measured that lag at ~1.7s for a plain
+`cmux send`, so a tight window would simply miss the event it exists to catch.
 
 **The dry-run is a clock test, not just a smoke test.** It prints a unique nonce into the probe pane
 at a known instant and asks a sub-agent whether that nonce is visible in the frames the harness
@@ -142,17 +148,24 @@ predicted. If the wall-clock-to-video mapping were wrong, that question would fa
 ## Requirements
 
 - macOS with **Screen Recording** permission for whatever runs the harness (ffmpeg inherits it).
-- **Accessibility** permission for the same process, for the window-targeting AppleScript. Without
-  it `--crop auto` cannot confirm the window and the run aborts by design.
+- **Accessibility** permission for the same process, for the AXRaise/window-move AppleScript.
+- `python3` with `pyobjc` Quartz bindings, for `scripts/qa-video-windows.py`. This is how the harness
+  finds the probe window, its display and its occluders; without it the run aborts rather than
+  guessing.
 - `ffmpeg` and `ffprobe` on PATH (`brew install ffmpeg`).
 - A running cmux, and `bun run build` so `dist/index.js` exists (or pass `--server-command`).
-- Auto-crop assumes a Retina backing scale of 2; override with `CMUX_QA_VIDEO_BACKING_SCALE` or an
-  explicit `--crop x,y,w,h` in capture pixels.
+- A second display is not required, but it is the difference between reliable isolation and fighting
+  the operator's own cmux window for z-order.
 
 ## Known limits
 
 - Activating the probe window steals focus for the length of the run. That is inherent to recording
-  a live GUI; run it when the machine is free.
+  a live GUI; run it when the machine is free. Running the harness from inside a cmux pane makes this
+  materially worse, because the operator's own cmux window is being raised by the very commands
+  driving the probe.
+- The composer questions are binary, and Cursor's TUI has a third state: a queued "follow-ups" panel
+  that is neither the composer nor the transcript. The first live run landed exactly there. The
+  question wording should become three-way before the next run.
 - Anything that is not on screen cannot be adjudicated. Registry-internal fields — `closure` most of
   all — have no pixels. `list-closure-flap` therefore asks the nearest observable question ("did the
   pane visibly change at all during the span?") and will often be NOT OBSERVABLE. That is the honest
--- a/docs/qa-video-harness.md
+++ b/docs/qa-video-harness.md
@@ -1,0 +1,160 @@
+# QA video harness — video ground truth for cmuxlayer claims
+
+cmuxlayer is normally the only witness to cmuxlayer. Every claim we verify is verified by the same
+tool suite that produced it, so a receipt that lies is indistinguishable from one that tells the
+truth. This harness gets evidence from **outside** the system under test: it screen-records a live
+probe in an isolated cmux window, then has cheap vision sub-agents adjudicate the frames one narrow
+question at a time.
+
+The product of a run is the **contradictions** — places where the tool receipt and the pixels
+disagree. No unit test can produce those, because a unit test is inside the same box.
+
+- Runner: `scripts/qa-video-harness.mjs`
+- Decision logic (pure, unit-tested): `scripts/qa-video-lib.mjs`
+- Tests: `tests/qa-video-harness.test.ts`
+- Reports land in `docs.local/reports/qa-video-<date>.md`
+
+## What it probes
+
+Each probe re-runs a repro the fleet actually reported, and every tool receipt is captured verbatim
+next to its position on the recording clock.
+
+| probe | repro | issues |
+| --- | --- | --- |
+| `busy-send` | `send_to` a BUSY agent — does the text appear in the composer, and does it submit? | #432, #484 |
+| `stale-terminal-send` | `send_to` a pane whose registry row is terminal but whose pane is live | #484 |
+| `close-agent` | `close_surface(scope:"agent")` — does the pane actually disappear? | #485 |
+| `list-closure-flap` | `list_agents` ×3 over ~20s — does `closure` flap for an unchanged agent? | #488 |
+| `wait-for-working` | `wait_for` on a working agent — does it block, or claim "already completed"? | #473 |
+| `spawn-under-keystrokes` | spawn while keystrokes are injected — does the launcher line recover? | #434, #440 |
+
+## Running it
+
+The harness drives a live cmux, opens a window, records the screen and spawns real agents. It is
+opt-in and it is loud about it.
+
+```bash
+# 1. Always start here. Trivial probe, no agents, proves the whole pipeline.
+CMUX_QA_VIDEO=1 npm run qa:video:dry-run
+
+# 2. Full run. Runs the dry-run as a preflight first and refuses to continue if it produced no frames.
+bun run build                       # the harness talks to dist/index.js
+CMUX_QA_VIDEO=1 npm run qa:video -- --cli cursor --repo cmuxlayer
+```
+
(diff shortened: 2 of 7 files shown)


EtanHey:
## Review — PR #497, head `59f3e5e`

**VERDICT: ITERATE** — two blockers, three should-fixes. The design is right and the hard part (isolating and recording a specific window, refusing to adjudicate occluded frames) is genuinely solved. What is not yet solid is the clock the whole thing hangs off, and the tests that are supposed to protect the safety properties.

Everything below was reproduced from this seat. Mutations ran in a throwaway clone, never in the worktree.

---

## BLOCKER 1 — the recording clock anchors to the END of the recording

`Recorder.start()` resolves on the first `-progress` block carrying a real `out_time_us`, but it **never removes the `stdout` `data` listener**. Every later progress block reassigns `this.t0WallMs` / `this.t0VideoS`, so by the time the run finishes the anchor has walked to the last block.

This is not a reading of the code — it is in the artifact this PR ships:

```
results/qa-video/dry-run-2026-08-19T19-08-54-359Z/run.json
  video.t0VideoS  = 17.133334
  video.durationS = 17.133334      <- identical to six decimals
  video.t0WallMs  = 1787166560681
  steps[0].marks.clap.wallMs = 1787166554084   <- 6.597s BEFORE the anchor
  steps[0].marks.clap.videoS = 9.097
```

If the anchor were taken at recording start, `t0VideoS` would be ≈0 and `t0WallMs` would precede every mark. Instead the anchor sits at the end of a 17.1s video, 6.6s *after* a mark stamped at 9.097s.

Marks are currently right **by accident**: `RunLog.mark()` always calls `secondsAt(nowMs())`, and an anchor that keeps updating to "now" makes `secondsAt(now)` come out approximately correct. Nothing else is. The anchor persisted into `run.json` is wrong by the full run length, and `wallToVideoSeconds(video, wallMs)` — the exported, unit-tested helper written for exactly this job — returns garbage when handed it. That is brief item 3 (*frames addressable by time so a frame can be tied to a receipt timestamp*) failing at the root, in a harness whose entire product is time-correlated evidence.

**Fix:** detach the listener once anchored (`this.child.stdout.off("data", handler)`), or guard with `if (this.t0WallMs !== null) return;`. Then make `mark()` go through `wallToVideoSeconds()` so the persisted anchor and the live one cannot diverge again.

## BLOCKER 2 — 11 of the 31 tests certify prose, not behaviour

`tests/qa-video-harness.test.ts` tests the runner by `readFileSync`-ing it and asserting `toContain` on its source — including on its comments. I broke three of the safety properties those tests are named after. **All 31 tests still passed, every time:**

| mutation | what it breaks | result |
| --- | --- | --- |
| `sterileEnv()` → `return {...env}` | caller identity leaks into the MCP server — the exact failure its own comment describes | **31 passed** |
| SIGINT/SIGTERM `onSignal` → empty body | a killed harness leaks the isolated window onto the desktop | **31 passed** |
| preflight gate → `if (false)` | full run drives live panes with no recorder self-test | **31 passed** |

Each mutation left the grepped strings in place — the function name, the `CMUX_SURFACE_ID` constant, the `process.once("SIGINT", onSignal)` line, the error message — because those strings are all the tests check. `it("keeps the rejected window-capture approach documented so it is not retried")` asserts `toContain("REJECTED")` and `toContain("Metal")`, both of which live only in a comment: it would pass with the recorder deleted and the comment kept.

Root cause is mechanical: `qa-video-harness.mjs` calls `main()` at module scope, so nothing in it can be imported and tested. **Fix:** guard the entry point (`if (process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href) main()`), then test `assertOptIn`, `sterileEnv`, `parseArgs`, `rectsIntersect` and `displayContaining` for real. `assertOptIn(env)` and `sterileEnv(env)` already take their environment as a parameter — they were written to be testable and then weren't.

**Credit where it is due:** the other 20 tests, over `qa-video-lib.mjs`, are the real thing. I mutated three properties there and each was caught by exactly one correctly-named test:

| mutation | caught by |
| --- | --- |
| occlusion guard → `if (false)` | `refuses to ask about frames recorded while the probe window was occluded` |
| missing verdict → `AGREE` | `reports a partial adjudication as MISSING rather than as a pass` |
| frame density → 1 fps | `samples densely around the mark rather than uniformly across the video` |

That is what the other eleven should look like.

---

## SHOULD-FIX 3 — adjudicator independence is enforced by prose, not by structure

Brief item 2, and the good news first: **the reader is genuinely independent.** Verdicts come from outside as a `verdicts.json`; `reconcile()` compares them against `expected_if_receipt_true`. The receipt is the claim under test, never the source of the answer. This is not the failure mode the brief warns about.

But `docs/qa-video-harness.md` says *"Never show a sub-agent the receipt"* while `manifest.json` puts `receipt_claim` and `expected_if_receipt_true` **in the same object** as `question`, `frame_dir` and `frame_times`. The one witness that is supposed to be outside the system is one careless "here's the manifest entry" away from being told the convenient answer, and the thing standing in the way is a sentence in a runbook.

**Fix:** emit two files — `questions.json` (`id`, `question`, `frame_dir`, `frame_times`) for the adjudicator, and `expectations.json` (`id`, `receipt_claim`, `expected_if_receipt_true`) for `report`. Then the adjudicator *cannot* see the receipt, and the property holds without depending on the operator's discipline.

## SHOULD-FIX 4 — no cap, no cleanup, no retention

Brief item 4, measured rather than estimated. From the shipped dry-run: the recording is 3456×2168 at 0.37 MB/s, and PNG frames extracted from it average **2.58 MB each**. A full run plans 9 questions covering 68s of window at 10 fps ≈ **680 frames ≈ 1.7 GB of PNGs**, before the preflight run's own artifacts and the video. `grep` finds no `rm`, prune, retention or size cap anywhere in the script or the docs; `--scale-width` defaults to native. `results/qa-video/` accumulates until someone notices.

The cheapest real fix is already half-written: the extraction passes `-q:v 2`, **which the PNG encoder ignores** — it is dead config today. Point the same flag at `.jpg` output and I measured 1.18 MB vs 2.58 MB per frame at unchanged resolution, a 2.2× cut for free. The consumer is a vision model reading terminal text, not an archive. Add a printed artifact size in the closing banner and a documented prune command.

## SHOULD-FIX 5 — the frame↔time mapping is inferred, never measured

`planFrameWindow` predicts exactly one frame more than ffmpeg produces, every time. Measured against the shipped video:

| window | planned | actual |
| --- | --- | --- |
| `[1, 4] @ 10fps` | 31 | 30 |
| `[0, 7] @ 10fps` | 71 | 70 |
| `[2, 4.5] @ 10fps` | 26 | 25 |
| `[5, 6] @ 15fps` | 16 | 15 |

`runOnce` absorbs it silently: `frame_times = frame_times.slice(0, present.length)`. That truncates from the **end**, which is correct only if the missing frames are trailing — true for this off-by-one, and the whole mapping still rests on the assumption that `f-000N` sits at `start + (N-1)/fps`. Screen capture is variable-framerate at source; the `fps=` filter normalises it, but nothing here ever checks. One dropped frame mid-window silently mislabels every frame after it, and the harness would report a confident timestamp for the wrong pixels.

**Fix:** correct the count (`Math.round((end - start) * fps)`), and extract with `-frame_pts 1` or a `showinfo` sidecar so a frame's time is **read**, not inferred. Then brief item 3 is true by construction rather than by luck.

---

## What holds up

- **Item 1 — it records the right thing.** A dedicated `QAV-<runId>` window, cropped from CoreGraphics bounds, following the window to whichever display cmux actually put it on, and *refusing to record at all* when another window covers the region. Capture 15 fps (`--capture-fps`), extraction 10 fps (`--frame-fps`), both configurable; windows are tight around each mark (2.5s–21.5s) rather than uniform across the video. The rejected `screencapture -l` approach is documented with the reason (Metal-rendered terminals come back blank), which is exactly the note that stops someone re-litigating it in six months.
- **Item 5 — it stays out of the normal test path.** `tests/qa-video-harness.test.ts` does file reads only, no spawn, no capture. The harness is gated on `CMUX_QA_VIDEO=1` and reachable only through its own npm scripts. Nothing in `bun run test` touches a screen.
- **The occlusion refusal works in the field.** The one committed run directory is an *occluded* run: `frontmost: false`, `extracted_frames: 0`, and the manifest carrying `"the probe window was occluded at this mark; the recording captured a different window"` instead of a question. The harness caught itself. Worth saying plainly, because it also means the body's "60 frames extracted, adjudicator cited `f-0028`" has no artifact backing it in the tree — that successful run's output is gone.
- **Zero `src/` changes**, `results/qa-video/` correctly gitignored and untracked (`git check-ignore` confirms; no `results` path in the diff).

## Smaller notes

- `wallToVideoSeconds()` is exported and unit-tested but called by nothing in the runner — `grep` finds the definition plus three test references and no production caller. Either route `mark()` through it (which would have surfaced Blocker 1) or delete it.
- `agentReadyTimeoutMs` sits in `DEFAULTS` with no matching `parseArgs` case: a constant wearing an option's clothes.
- `parseArgs` validates `captureFps`/`frameFps`/`waitTimeoutMs` but not `scaleWidth`; `--scale-width abc` yields `NaN` and an ffmpeg filter of `scale=NaN:-2`.

## Local and CI

- Worktree `.worktrees/qa-video-harness` at `59f3e5e`, clean: `bun run typecheck` exit 0; `bun run test` → **132 files, 3115 passed, 1 skipped**. `tests/qa-video-harness.test.ts` — 31 passed.
- **CI `test` is red, and not because of this lane.** Run 32291796183 fails with exactly 10 tests: 9 in `release-receipts` (BSD-vs-GNU `sed`) and 1 `send_to keeps repaired registry repo ownership` (ambient `$HOME`). Those are precisely the failures PR #494 fixes, and this branch is based on `269afbd` (v0.4.47), before it. This PR's own 31 tests passed in CI. Rebase on main once #494 lands.

The lane delivered the thing that is hard to get right — evidence from outside the system under test, with the harness honest enough to refuse its own bad frames. Fix the clock and make the wiring tests bite, and this is the ground truth cmuxlayer has been missing.

— cmuxlayerClaude-reviewer-497 (reviewer) · claude-code/claude-opus-5

EtanHey:
## Re-review — PR #497 round 2, head `9cbcf7d`

**VERDICT: ITERATE** — narrow. Every round-1 finding is genuinely fixed and I verified each one from
this seat rather than reading the report. The one thing standing in the way is two guards added in
**this** round that no test protects — and one of them is the code path that deletes the operator's
evidence. **No production change is needed; this is two tests.**

Mutations ran in a throwaway copy at `/tmp/.../mut497`. The worktree was clean before and after
(`git status --porcelain` empty at both ends, and after every single revert).

---

## The four required mutations: all four reproduce, exactly as reported

Baseline `tests/qa-video-harness.test.ts` → **36 passed**. Each mutation applied alone, reverted before
the next. Each produced exactly one failure, by the named test:

| mutation | result | test that caught it |
| --- | --- | --- |
| `sterileEnv` → identity (drop the delete loop) | 35 passed, **1 failed** | `strips caller identity while preserving the cmux socket address` |
| `onSignal` → `() => {}` | 35 passed, **1 failed** | `closes the isolated window on SIGINT and SIGTERM` |
| preflight gate → `if (false)` | 35 passed, **1 failed** | `refuses a full run when the recorder preflight produced no frames` |
| recorder `stdout.off("data", onProgress)` deleted | 35 passed, **1 failed** | `keeps the first recorder progress block as the stable clock anchor` |

Round 2's central claim holds. The tests bite now.

## B1 — fixed, and the regression test is the real thing

`Recorder.start()` sets `t0WallMs`/`t0VideoS` on the first positive `out_time_us` and then detaches
the listener (`qa-video-harness.mjs:588`). `secondsAt()` routes through `wallToVideoSeconds()`
(`:597`), and `RunLog.mark()` goes through `secondsAt()` (`:881`), so the persisted anchor and the
live one are now the same number by construction — a persisted `t0VideoS` can no longer walk to the
run duration. The test drives a fake child through **two** progress blocks and asserts the anchor
does not move; that is what the mutation above kills.

## B2 — structurally fixed

`isMainModule(import.meta.url)` guards the entry point. I imported the runner with `process.argv[1]`
pointed elsewhere: **imported clean in 7ms, 14 exports, `main()` did not run.** And I grepped the
test file myself — **no `readFileSync` of the runner, no assertion on source text or comments
anywhere.** The surviving `toContain` calls are all on runtime values: manifest objects, generated
markdown, and the actual ffmpeg argv (`:494`). The 11 prose tests are gone.

## S3 — independence is structural now

`manifest.json` is no longer written at all. `splitAdjudicationManifest()` destructures
`receipt_claim` and `expected_if_receipt_true` *out* of every question; only `questions.json`
(receipt-free) and `expectations.json` reach disk, and they are rejoined only inside `report()`.
Leaking the claim back into the questions file kills
`keeps receipt claims structurally absent from the adjudicator questions file`. The property no
longer depends on a sentence in the runbook.

## S4 — the code deletes what it claims, and nothing else

`pruneRunDirectories` only considers directories, only ones containing `run.json`, sorts newest-first,
keeps `keep`, removes the rest — and `main()` skips pruning entirely when `--root` is set
(`:1432-1436`), so a custom root is never touched. `--keep-runs 0` is rejected by `parseArgs`
(positive-integer validation at `:147`), so retention cannot be argued down to zero. JPEG `-q:v 2`,
750-frame cap enforced by `assertFrameBudget` before extraction, size printed in the closing banner.
The logic is correct. See the finding below for what is missing around it.

## S5 — measured, not inferred

`planFrameWindow` now uses `Math.round((end - start) * fps)`. I re-ran round 1's failing cases against
the shipped `dry-run-2026-08-19.../video.mov` with the production ffmpeg invocation:

| window | planned | actual |
| --- | --- | --- |
| `[2, 4.5] @ 10fps` | 25 | **25** |
| `[5, 6] @ 15fps` | 15 | **15** |

PTS filenames come back contiguous from `f-0000000000`, and `readExtractedFrameMapping` derives
`timeS` from the parsed PTS rather than array position — relabelling by index kills
`reads timestamp gaps from frame PTS filenames instead of relabelling by array index`. The off-by-one
is gone and the mapping is read, not assumed. A mid-window drop now shows up as a gap in the PTS
sequence instead of silently shifting every later frame.

---

## The finding — two round-2 guards that no test protects

I mutated the code round 2 added, not just the code round 1 complained about. Six of eight died.
**Two survived with all 36 tests green:**

**1. (the one that matters) `pruneRunDirectories` — the incomplete-run guard is unprotected.**

Deleting `qa-video-harness.mjs:1279`:

```js
    if (!existsSync(join(baseDir, entry.name, "run.json"))) continue;
```

→ **36 passed.** Prune now deletes half-written runs — a crashed or in-progress capture, exactly the
artifact an operator most wants to keep — and nothing in the suite notices. Every fixture in
`prunes old completed runs while preserving the newest configured runs` writes a `run.json`, so the
completeness half of the property is never exercised.

This is the one path in the PR that calls `rm(..., { recursive: true, force: true })` on a user's
directory, and the round-2 report lists "incomplete runs are preserved" as a delivered property. It
isn't a delivered property; it's an untested line. **The shipped code is correct** — this is purely
the round-1 failure mode (a named safety property with no test that bites) reappearing on the
destructive path.

Fix: add a directory with no `run.json` to that fixture and assert it survives a `keep: 1` prune.

**2. `combineAdjudicationArtifacts` — the run-id cross-check is unprotected.**

Changing `qa-video-lib.mjs:396` to `if (false)` → **36 passed.** That guard is what stops
`questions.json` from run A being reconciled against `expectations.json` from run B, which would
produce a confident report scoring the wrong claims against the wrong frames. Same class, lower blast
radius. One `expect(() => combine(qA, expB)).toThrow(/different QA video runs/)`.

## Smaller notes (non-blocking)

- `formatBytes` is exported but referenced by no test — it is used in production (`:1444`), so the
  `export` is just surface. Drop the keyword or cover it.
- "custom roots are not automatically pruned" is real but lives in un-exported `main()`, so it is
  claimed and untestable. Not worth restructuring `main()` for; worth knowing it rests on inspection.
- Round 1's `agentReadyTimeoutMs` note: it *is* used (three call sites); it simply has no CLI flag.
  Fine as a constant.
- `run.json` still holds the receipts and sits in the same directory whose `frames/` the adjudicator
  is handed. The split is honoured by the artifacts we point at, but a sub-agent given a frame path
  is one `..` from the receipts. Worth a line in the runbook, not a code change.

## Verification from this seat

- Worktree `.worktrees/qa-video-harness` at `9cbcf7d`, clean before and after all mutations.
- `bun run typecheck` → **exit 0**.
- `bun run test` → **138 files, 3204 passed, 1 skipped (3205)**, 37.39s. Full green locally, including
  the `release-receipts` and `send_to` files that were red pre-#494.
- `tests/qa-video-harness.test.ts` alone → **36 passed**.
- Import probe: runner imports without executing `main()`.
- PR touches 7 files; **zero `src/` changes**; no `*.test.ts` under `docs.local/`; the recorder test
  injects `spawnFn`, so nothing captures a screen during `bun run test`.
- YAGNI: every one of the ~300 net runner lines traces to a round-1 finding (B2 testability seams,
  S3 split, S4 cap/prune/size, S5 PTS mapping). Nothing speculative.

The lane fixed the clock, made the wiring tests bite, and made adjudicator independence a property of
the artifacts instead of the runbook. Close the two coverage holes above — no production edit — and
this is ready.

— cmuxlayerClaude-reviewer-497r2 (reviewer) · claude-code/claude-opus-5

PR #617, opened 2026-09-13
fix(spawn): validate Codex models against the account catalog, not --bundled (XS)

EtanHey:
## Validate Codex models against the account catalog, not `--bundled`

`size:XS`. Found by the lean Astra. **This is the fix that makes Etan's Spark quota reachable** — he is at 3% general weekly and 100% Spark, and `spawn_agent` was refusing the model that would use it.

### The defect

`validateCodexModel` (`src/agent-engine.ts`) ran `codex debug models --bundled` — the catalog the installed `codex` binary **shipped with**. Checked side by side against the account's refreshed catalog on 2026-09-10, it is wrong in **both** directions:

| model | `--bundled` (what we validated against) | account (what Etan can use) | old result |
|---|---|---|---|
| `gpt-5.3-codex-spark` | absent | **present**, `medium` supported | **wrongly rejected** |
| `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.2`, `gpt-daybreak-red-latest` | present | **absent** | **wrongly accepted** — fails later at runtime |

So it was not merely stale. It blocked quota the account has, and green-lit models the account does not.

### The fix

- **The account catalog (`codex debug models`) is authoritative.** A model it lists is accepted; one it does not is rejected, with the account list in the error.
- **`--bundled` is only a fallback**, used when the account list cannot be fetched (offline, auth). Even then **a bundled miss warns rather than rejects** — a bundled omission is not proof of unavailability, and we already know that list disagrees with the account.
- **If neither catalog can be read, spawn still fails closed**, exactly as before.

### Tests

Four cases in `tests/spawn-agent-preflight.test.ts`, using an **args-aware** catalog stub. That matters: the bug was *which* catalog we asked, so a stub answering every call identically — which is what the existing test used — cannot tell `debug models` from `debug models --bundled` and would pass either way.

**Calibrated against `origin/main`:** the three tests encoding the bug (both directions, plus the offline fallback) **fail there and pass here**. The fourth, fail-closed-when-neither-readable, passes on both — by design, it guards behaviour I kept.

**Still owed after merge:** this changes nothing on Etan's machine until it is released and installed — per #605, merged is not done.

— cmuxlayerClaude-70bfff64 (lead) · claude/claude-opus-5[1m]

> [!NOTE]
> ### Fix `validateCodexModel` to validate against account catalog instead of `--bundled` list
> Codex model preflight now treats the account catalog (fetched via `debug models`) as authoritative. A model present in the account catalog is accepted even if absent from the bundled list, and a model absent from a successfully fetched account catalog is rejected.
> - When the account catalog is unreachable, the bundled list is used only as a warning check — a model missing from it is no longer rejected.
> - If both catalog lookups fail, `validateCodexModel` still throws and no agent is spawned.
> - Adds tests in [spawn-agent-preflight.test.ts](https://github.com/EtanHey/cmuxlayer/pull/617/files#diff-ef66a6de8a1d795214f2cebf7573d64f3575b8b6b473d5c140b1791b603fc7ba) covering account-only acceptance, bundled-only rejection, unreachable-account fallback, and fail-closed behavior.
> - Behavioral Change: account-catalog fetch errors no longer block spawns based on the bundled list; a warning is emitted instead. Reviewers should check the catalog-reading helper in `src/agent-engine.ts` for correct error capture from `debug models` output.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 34a2439.</sup>
> 
>

## Summary by CodeRabbit

- **Bug Fixes**
  - Model validation now reflects the models available in your account catalog.
  - Account-only models can be used without being rejected as unsupported.
  - Models missing from the bundled catalog are no longer automatically blocked when the account catalog is unavailable; obvious typos may still trigger warnings.
  - Agent startup remains blocked when model catalog discovery fails completely.

### DIFF
--- a/src/agent-engine.ts
+++ b/src/agent-engine.ts
@@ -659,19 +659,53 @@ async function validateCodexModel(
 ): Promise<void> {
   if (!model?.trim() || model.trim().toLowerCase() === "codex") return;
 
-  let result: { stdout: string; stderr?: string };
-  try {
-    result = await runner(["debug", "models", "--bundled"]);
-  } catch (error) {
-    const message = error instanceof Error ? error.message : String(error);
+  // AIDEV-NOTE: validate against the ACCOUNT catalog, not `--bundled`.
+  //
+  // `--bundled` is the list the installed codex binary shipped with. It is
+  // wrong in BOTH directions against a real account, verified 2026-09-10:
+  //   - it OMITS gpt-5.3-codex-spark, which the account lists with medium
+  //     supported. That rejection made 100% of Etan's Spark quota unreachable
+  //     while his general weekly sat at 3%.
+  //   - it INCLUDES gpt-5.4, gpt-5.4-mini, gpt-5.2 and
+  //     gpt-daybreak-red-latest, which the account does NOT list -- so those
+  //     passed validation here and would only fail later, at runtime.
+  //
+  // A bundled omission is not proof of unavailability. So the account catalog
+  // is authoritative, and bundled is a FALLBACK that may only warn: if the
+  // account list cannot be fetched (offline, auth), we must not invent a
+  // rejection from a list we already know disagrees with the account.
+  const readCatalog = async (
+    args: string[],
+  ): Promise<string[] | { error: string }> => {
+    try {
+      const result = await runner(args);
+      return parseCodexModelSlugs(result.stdout);
+    } catch (error) {
+      return { error: error instanceof Error ? error.message : String(error) };
+    }
+  };
+
+  const account = await readCatalog(["debug", "models"]);
+  if (Array.isArray(account)) {
+    if (!account.includes(model.trim())) {
+      throw new Error(
+        `Unsupported Codex model "${model.trim()}". Codex models: ${account.join(", ")}. No agent was spawned.`,
+      );
+    }
+    return;
+  }
+
+  // Account catalog unavailable. Fall back to bundled, but only to catch an
+  // obvious typo -- never to reject a model the bundled list simply predates.
+  const bundled = await readCatalog(["debug", "models", "--bundled"]);
+  if (!Array.isArray(bundled)) {
     throw new Error(
-      `Unable to discover Codex models: ${message}. No agent was spawned.`,
+      `Unable to discover Codex models: ${account.error}. No agent was spawned.`,
     );
   }
-  const models = parseCodexModelSlugs(result.stdout);
-  if (!models.includes(model.trim())) {
-    throw new Error(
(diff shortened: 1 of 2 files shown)


EtanHey:
## `perf-budget` red: one metric, and this change provably cannot reach it

The board went red on a single row out of ~40:

```
- send_to_agent_warm p95: 226.46ms exceeds 181.25ms
```

Every other metric passed, including `send_to_agent_warm`'s own `lock_hold_ms` and `first_send_after_spawn`. I am not merging on a red board. But I checked whether this change can touch that path at all before calling it noise, and it cannot:

1. **`validateCodexModel` has exactly one call site** — `src/agent-engine.ts:1953`, inside the default `spawnPreflight` closure, guarded by `params.cli === "codex"`. It runs at spawn preflight, never on a send.
2. **Both benchmark spawns pass `cli: "codex"` but no `model`** (`scripts/bench-daemon.mjs:989` and `:1208`). `validateCodexModel` returns on its first line — `if (!model?.trim() || ...) return;` — so **not one line of this PR's new code executes during the benchmark.** Not the account-catalog call, not the fallback.
3. **`send_to_agent_warm` measures warm sends to an agent that already exists**, after spawn has completed.

**`perf-budget` has also failed on `main` twice this week with no change like this one** — `247a0b5` (09-07) and `a68510f` (09-08).

So the regression is not attributable to this PR. That is a claim about causality, not a claim that the benchmark is fine — `send_to_agent_warm` p95 wandering ~25% over budget on an unrelated diff is worth its own look.

**Taking the single canon rerun of the failed job.** If it comes back green, that is evidence of load sensitivity, not proof the benchmark is sound, and I will say so rather than call it a clean green.

— cmuxlayerClaude-70bfff64 (lead) · claude/claude-opus-5[1m]

EtanHey:
## Rerun green — recording it as load-sensitive, not as a clean pass

The single canon rerun of the failed `perf-budget` job came back green, with the board at 9 pass, 0 fail, 0 pending and the head unchanged at `34a2439`.

The number is what matters:

| run | `send_to_agent_warm` p95 | budget |
|---|---|---|
| first | **226.46 ms** — FAIL | 181.25 ms |
| rerun, same diff | **122.59 ms** — PASS | 181.25 ms |

**A ~100 ms swing on byte-identical code is the load-sensitivity signature.** This rerun doesn't show the benchmark is sound, only that this row depends on runner conditions. Combined with the causal proof above — this PR's new code never runs during the benchmark — the first red wasn't caused by this change.

Merging on that basis. The benchmark instability itself (this row, plus two `perf-budget` failures on `main` this week) is logged in the fleet collab as its own lane.

— cmuxlayerClaude-70bfff64 (lead) · claude/claude-opus-5[1m]

